PR #24, opened 2015-08-07
Building ALFA with Intel compilers

theWayofthecode:
I selected the following options after executing
./alfaconfig.sh

2) Intel Compiler (Linux)
1) No Debug Info
2) No (to Root 6)
1) Yes (to simulation engines)
2) Internet
2) No (to Python bindings)

My environment details are.
icc version 15.0.2 (gcc version 4.9.0 compatibility)
Scientific Linux CERN SLC release 6.7 (Carbon) kernel 2.6.32-431.el6.x86_64

Problems I faced:
- Initially my compiler wouldn't be detected because there wasn't a test case for intel (see first git commit)
- For the boost installation I would get "boost installation error: Unknown toolset: intel". The second commit fixed the problem.
- For root, although there wasn't any compilation errors, the software would be compiled 10 times an d then it would terminate with an error message that the 10 tries were exhausted. I checked the workaround in the install_root6.sh script and I realized that a make install was missing, therefore the until expression wouldn't become true. IMO we don't need the workaround anymore. In the future If the compiler crashes (due to an internal bug) we can temporarily disable support for icc, and file a bug to the compiler maintainers.

In the end I succeeded to build the whole thing with intel compilers, after I solved few other issues along the way.


fuhlig1:
Merged with commit e317aaa0ef9aa94cea5841a4daeb647cbc96c672

PR #56, opened 2017-03-08
New CMake required for Root6

janmayer:
On a system with outdated cmake, installing root6 fails with 
```
CMake Error at CMakeLists.txt:2 (cmake_minimum_required):
  CMake 3.4.3 or higher is required.  You are running version 3.3.2
```
The cmake version installed by FairSoft is too old as well. Require 3.4.3 and install 3.7.2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ directory of FairSoft.
 
 ##Included Packages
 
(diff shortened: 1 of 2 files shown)


MohammadAlTurany:
Solved with other issues in PR #64

PR #60, opened 2017-07-25
Adding JSROOT (Java Script ROOT)

DanielWielanek:
Adding JSROOOT to FairSoft framework.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -354,6 +354,16 @@ then
   source scripts/install_msgpack.sh
 fi
 
+
+##################### JsROOT  ##################################################
+
+if [ "$check" = "1" ];
+then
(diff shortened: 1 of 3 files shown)


fuhlig1:
Hi Daniel,

sorry for the very late response but I am not checking the issues on github and I did not receive a notification mail.

I am not sure if adding of jsroot makes sense. You get a version of jsroot with your Root installation even if this version is normally rather old. With your PR you would get a slightly newer version, but this version will very soon also be outdated. In my opinion it is better to download  jsroot in the requested version when needed. With a simple definition of one environment variable this  can be done by any user such that any user. 

What do you think.

Ciao

Florian

PR #70, opened 2018-04-23
Bump ZeroMQ and DDS versions

dennisklein:
Build ZeroMQ with CMake. The bumped DDS version contains a fix for
installing DDS in the same prefix with Boost.

---

@MohammadAlTurany You can find the equivalent for aliBuild in FairRootGroup/alfadist#1.

### DIFF
--- a/scripts/install_zeromq.sh
+++ b/scripts/install_zeromq.sh
@@ -21,7 +21,7 @@ then
   cd build
 
   cmake .. \
-        -DWITH_PERF_TOOL=OFF \
+        -DWITH_PERF_TOOL=ON \
         -DZMQ_BUILD_TESTS=ON \
         -DENABLE_CPACK=OFF \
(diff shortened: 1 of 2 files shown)


scripts/install_zeromq.sh
```diff
@@ -1,59 +1,46 @@
...
+  git checkout $ZEROMQ_VERSION
+  if [ ! -d  build ];
+  then
+    mkdir build
+  fi
+  cd build
 
-    check_all_libraries  $install_prefix/lib
+  cmake .. \
+        -DWITH_PERF_TOOL=OFF \
```
rbx: Let's turn this on, the perf tools are useful sometimes to compare performance and the build times are not very high.

dennisklein: @rbx Done

PR #71, opened 2018-04-24
Add MQ_Only option to build only FairMQ with dependancies

MohammadAlTurany:


MohammadAlTurany:
Ok, Thanks. It is added it back.

PR #76, opened 2018-04-26
Built with builtin libcurl

dennisklein:
The way the system libcurl is used prohibits the correct
evaluation of the NO_PROXY environment variable.

---

As a reminder for the details see the following discussion: https://gitlab.kitware.com/cmake/cmake/merge_requests/1409

### DIFF
--- a/scripts/install_cmake.sh
+++ b/scripts/install_cmake.sh
@@ -75,7 +75,7 @@ then
   then
     cd $SIMPATH/basics/cmake
     ./bootstrap --prefix=$install_prefix --docdir=/share/doc/CMake --mandir=/share/man \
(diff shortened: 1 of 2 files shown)


dennisklein:
Yes, tested on lxbk0199 and cmake 3.9.4 built with `scripts/install_cmake.sh` without the flag.

Side note: I only made sure the builtin libcurl is new enough from cmake 3.9+. Once the system libcurl version caught up we can use it again, because the bug was fixed in libcurl and is only visible, if you use libcurl the same way cmake does (combination of env variable and easy C interface of libcurl). With the command line interface curl you cannot reproduce - there `NO_PROXY` always worked correctly.

PR #109, opened 2018-09-10
Fix Geant3 with newer gfortran versions

PMunkes:
Building Geant3 with gfortran versions > 8.1 will fail with hundreds of errors of type:
`Error: Actual argument contains too few elements for dummy argument ‘X’ (6/50) at (1)`
This is an intentional behaviour change:
> * When an actual argument contains too few elements for a dummy argument, an error is now issued. The -std=legacy option can be  used to still compile such code.
[Source](https://www.gnu.org/software/gcc/gcc-8/changes.html)

This pull request fixes this issue with (hopefully) minimum side effects. This patch series could also be backported to older FairSoft releases.

### DIFF
--- a/scripts/install_geant3.sh
+++ b/scripts/install_geant3.sh
@@ -21,6 +21,8 @@ then
 
   cd $SIMPATH/transport/geant3
 
+  mypatch ../geant3_gfortran.patch | tee -a $logfile
+  
 # ????
 #  if [ "$build_root6" = "yes" ]; then
 #    mypatch ../geant3_root6.patch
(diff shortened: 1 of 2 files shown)


fuhlig1:
Hi @MohammadAlTurany,

shouldn’t we better take Geant3 from our own repository  were the issue is already fixed. This would also cleanup the installation script since all patches are now included in our Geant3 repository.

MohammadAlTurany:
Hi @fuhlig1 
Sure. Please do it, reference it in this PR and close it.
best,
Mohammad

PMunkes:
Thanks for the swift reply, feel free to close this PR when it is no longer needed.

fuhlig1:
Decided against using Geant3 from our own repo. For oct17 we use Geant3 2.1 where there is no fixed version in our repo. Applied the patch as proposed.

PMunkes:
I agree that the patch looks strange, I already pushed a new version to my feature branch. I will open a merge request for this fixed version as soon as possible.

fuhlig1:
Hi @kgertsenberger,

the problem should be solved wit the new release oct17p3.

kgertsenberger:
Ok, thank you.
P.S. the versions of msgpack and flatbuffer packages in FairRoot don't also support the new GCC 8 compiler.

PR #110, opened 2018-09-10
Fix geant3 newer gfortran

PMunkes:
Building Geant3 with gfortran versions > 8.1 will fail with hundreds of errors of type:
`Error: Actual argument contains too few elements for dummy argument ‘X’ (6/50) at (1)`
This is an intentional behaviour change:
> * When an actual argument contains too few elements for a dummy argument, an error is now issued. The -std=legacy option can be  used to still compile such code.
[Source](https://www.gnu.org/software/gcc/gcc-8/changes.html)

This pull request fixes this issue with (hopefully) minimum side effects. This patch series is backported.


PMunkes:
Closed as I forgot to rebase the changes. Sorry about that.

PR #111, opened 2018-10-02
Fix geant4_vmc version in install script

PMunkes:
Geant4VMC was updated to version 3.6, but the linked directory in this file stayed at v3.3. 
This caused issues with several tests using the geant4libs.C macro, which could no longer be found.


PMunkes:
This patch could also be backported to the may18 release.

PMunkes:
Now with correct branch.

PR #115, opened 2018-11-13
Fix Build Issue with geant3 patch

PMunkes:
The patch in Issue #109 contains 3 superfluous lines at the beginning of the file that may cause the build to fail. This patch removes them.


fuhlig1:
Close the pull request.  A working version was introduced with af0b51. There is also a new patch release oct17p3

PR #128, opened 2019-03-04
Bump required and supplied CMake versions

rbx:
Supplied version - 3.13.4 (latest patch of a stable release).
Required version - 3.12.4 (latest patch of a release that supports boost 1.68).


fuhlig1:
Must also be applied to branch RC_feb19

PR #140, opened 2019-05-29
Deactivate asiofi on MacOS

rbx:
"Fixes" FairRootGroup/asiofi#5.


dennisklein:
I would also exclude `install_ofi.sh`.

rbx:
Right, deactivated ofi also.

PR #146, opened 2019-06-25
Bump DDS to 2.4

dennisklein:
Needed for alfa-ci environment `fairmq_dev`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ directory of FairSoft.
 | nanomsg |1.1.5|
 | FlatBuffers |1.11.0|
 | MessagePack |3.1.1|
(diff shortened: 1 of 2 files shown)


MohammadAlTurany:
@dennisklein  Should it also go the release candidate?

dennisklein:
> Should it also go the release candidate?

Only, if you plan to include FairMQ after `v1.4.4` which is not released yet. For me it is fine to stay in dev.

> you need to change the README.md too!

Done.

dennisklein:
> Should it also go the release candidate?

I guess it would also break some dds examples in FairRoot. DDS 2.4 changed the topology xml schema.

PR #154, opened 2019-09-03
Jun19 patches

kgertsenberger:
XRootD patch: installation of binary files and GCC 9 support


kgertsenberger:
P.S. I checked it with GCC 4.8.5, 7.4.0, 9.1.0

PR #158, opened 2019-10-08
Compile flatbuffers with -fPIC

rbx:


scripts/install_flatbuffers.sh
```diff
@@ -23,6 +23,7 @@ then
           -DCMAKE_BUILD_TYPE=Release \
           -DCMAKE_CXX_COMPILER=$CXX \
           -DCMAKE_C_COMPILER=$CC \
+          -DCMAKE_CXX_FLAGS="-fPIC" \
```
dennisklein:
```suggestion
          -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
```

If the above works, we should prefer it.

dennisklein: Just tried with FairMQ alfaci hosts and it appears to work.

rbx: Changed, thanks

PR #161, opened 2019-10-21
Fix ICU version check, rootconfig.sh OPENGL_gl_LIBRARY

klenze:
If one installs ICU 63.1 (package_versions.sh), then one should not expect ICU 53.1 to be installed afterwards.

### DIFF
--- a/scripts/install_icu.sh
+++ b/scripts/install_icu.sh
@@ -14,7 +14,7 @@ fi
 
 install_prefix=$SIMPATH_INSTALL
 
-checkfile=$install_prefix/lib/libicuuc.so.53.1
+checkfile=$install_prefix/lib/libicuuc.so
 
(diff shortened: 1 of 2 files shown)


fuhlig1:
Why do you need the fix? Are you building FairSoft on Linux with Clang?

klenze:
Yes, unfortunately. then I tried to compile root with clang9 on linux and could not get it to work, so this is probably moot now.

kresan:
Is it still an open issue @klenze ?
Meanwhile we migrate to the new build system - based on Spack. You can find it in the dev branch. Versions of jun19 are also available there - in env/jun19/* environments. Can you test if it solves your problem?

kresan:
There is no further patch release for jun19 planned. Closing this. Please use FairSoft based on Spack from the dev branch.

PR #176, opened 2020-01-27
DDS and ODC packages

AndreyLebedev:

PR #183, opened 2020-01-29
Update FairMQ and friends

### LINKED ISSUE #182: Fix git source caching upstream
Let's get spack/spack#14344 resolved.

dennisklein:
* [x] ~Needs~ Includes #176 
* [x] Needs #181 
* [x] Wants #178 
* [x] Make DDS `wn_bin` target work with non-system compilers
* Test
   * FairMQ
      * [x] Fedora31
      * [x] macOS10.14
      * [x] CI
   * DDS:
      * [x] Fedora31
      * [x] macOS10.14
      * [x] CI
   * ODC:
      * [x] Fedora31
      * [x] macOS10.14
      * [x] CI
   * FairLogger:
      * [x] Fedora31
      * [x] macOS10.14
      * [x] CI

---

* Cherry-picked @ChristianTackeGSI's FairLogger recipe from upstream.
* Transformed the touched packages to have git-based sources only and opted out of source caching (read in-line comment). Once we resolve #182 we get source caching back for those few packages (and this is orthogonal to this PR). All in all, I find a pure git-base packaging approach much simpler here - let's discuss.
* Cherry-picked also #176 ~, but will rebase on whatever gets merged finally.~ (included)


dennisklein:
@ChristianTackeGSI When loading a non-system compiler e.g. via a module and setting it up as a spack compiler, one has to set the `LD_LIBRARY_PATH` in the `compiler.yml` in order to build the DDS `wn_bin` target, e.g.:
```yaml
- compiler:
    spec: gcc@8.3.0
    paths:
      cc: /usr/lib64/ccache/gcc
      cxx: /usr/lib64/ccache/g++
      f77: /home/dklein/.spack/install_tree/linux-fedora31-skylake/gcc-9.2.1/gcc-8.3.0-xqpz6stvina7sfg3l7pm7lhscvtq57e6/bin/gfortran
      fc: /home/dklein/.spack/install_tree/linux-fedora31-skylake/gcc-9.2.1/gcc-8.3.0-xqpz6stvina7sfg3l7pm7lhscvtq57e6/bin/gfortran
    flags: {}
    operating_system: fedora31
    target: x86_64
    modules: []
    environment:
      prepend-path:
        LD_LIBRARY_PATH: "/home/dklein/.spack/install_tree/linux-fedora31-skylake/gcc-9.2.1/gcc-8.3.0-xqpz6stvina7sfg3l7pm7lhscvtq57e6/lib64:/home/dklein/.spack/install_tree/linux-fedora31-skylake/gcc-9.2.1/gcc-8.3.0-xqpz6stvina7sfg3l7pm7lhscvtq57e6/lib"
    extra_rpaths: []
```
Another way to pass through the `LD_LIBRARY_PATH` from the calling shell to the spack package is `spack install --dirty dds`.

dennisklein:
@ChristianTackeGSI 
https://github.com/FairRootGroup/FairSoft/blob/aeca4fd0516c2abb7929e44271056c56360a152b/packages/dds/package.py#L18-L27

* As we suspected, checkout from a branch is yet another case which leaves the source dir un`git describe`able. `get_full_repo=True` does the trick here.
* I also discovered the `no_cache=True` option which I find more expressive than using `branch=`. What do you think?

dennisklein:
@kresan This PR includes your `test_feb20` commit.

@ChristianTackeGSI @kresan On Linux I am quite happy with this. Now mac testing is missing and probably some adjustments on the fairroot recipe are needed - let's see. Reviews/Comments are welcome.

env/feb20/sim_python_no-threads.yaml
```diff
@@ -0,0 +1,21 @@
...
+  definitions:
+    - when: platform == 'linux'
+      root: [root@6.18.04+fortran+gdml+http+memstat+pythia6+pythia8+vc~vdt+python+tmva]
+    - when: platform == 'darwin'
+      root: [root@6.18.04+fortran+gdml+http+memstat+pythia6+pythia8+vc~vdt+python+tmva+aqua]
+  specs:
+    #    - cmake@3.13.4
+    - googletest@1.10.0
+    - boost@1.70.0
+    - fairlogger@1.6.1
```
dennisklein: > @dennisklein which versions of FairMQ & co will go into feb20 release? Can you add corresponding entries to the environment?

@kresan @MohammadAlTurany I would pin FairMQ, and DDS for jun19, because it was already defined in the legacy code. For new - spack-only - releases @ChristianTackeGSI and I remember both when we (all four of us) last had discussed this, we reached the conclusion to only pin versions of physics-sensitive packages, didn't we?

If you still agree with what I remember, I would suggest to remove the lines above:
```diff
@@ 8,4 - 8,0 @@
-    #    - cmake@3.13.4
-    - googletest@1.10.0
-    - boost@1.70.0
-    - fairlogger@1.6.1
```

Or did I miss an argument why we changed our opinions on this?

kresan: @dennisklein sorry, since our conclusion was not noted on the paper, I just forgot it. Agree, I will remove the lines you mentioned from the environment feb20.

ChristianTackeGSI:
I have issued a rerun of the CI job. This *should* have included the cmake fixes, because it should have done a merge with the base branch.

Can you please review the logs?
If the merge did not work out,
a) we should understand why
b) could you please rebase.

dennisklein:
> This should have included the cmake fixes, because it should have done a merge with the base branch.
Huh, how?

dennisklein:
FIXME: on witmac007:
```
*** Number of threads per python: test/buildsetup.sh: line 15: python3: command not found
```

ChristianTackeGSI:
Could we please move the CI related stuff into its own PR?
I have a few things to add there too. Including the above python3 thing.

dennisklein:
You can push to my branch.

packages/fairmq/package.py
```diff
@@ -9,74 +12,76 @@ class Fairmq(CMakePackage):
...
+            values=('11', '14', '17'),
+            multi=False,
+            description='Force the specified C++ standard when building.')
+    conflicts('cxxstd=11', when='@1.4.11:')
+
+    patch('fix_find_dds.patch', when='@1.4.0:1.4.4')
+
+    depends_on('googletest@1.7:', when='@:1.4.8')
+    depends_on('boost@1.64: +container+program_options+thread+system+filesystem+regex+date_time', when='@1.3')
+    depends_on('boost@1.64: +container+program_options+filesystem+date_time+regex', when='@1.4,develop')
```
ChristianTackeGSI: What about `when='@1.4:'`?
`develop` is larger than any version, so it's included in the open range.

packages/fairmq/package.py
```diff
@@ -9,74 +12,76 @@ class Fairmq(CMakePackage):
...
+    depends_on('fairlogger@1.2:1.5', when='@:1.4.7')
+    depends_on('fairlogger@1.2:', when='@1.4.8:,develop')
+    depends_on('zeromq@4.1.5:')
+    depends_on('nanomsg@1.1.5:')
+    depends_on('msgpack-c@3.1:')
+    depends_on('dds@2.2:2.4', when='@:1.4.9')
+    depends_on('dds@2.5-odc', when='@1.4.10')
+    depends_on('dds@3.0:', when='@1.4.11:,develop')
+    depends_on('flatbuffers', when='@1.4.9:,develop')
+    depends_on('pmix@2.1.4:', when='@1.4:,develop')
```
ChristianTackeGSI: Please drop all the `,develop`. They should already be included in the range.

packages/fairmq/package.py
```diff
@@ -9,74 +12,76 @@ class Fairmq(CMakePackage):
...
+
+    depends_on('cmake@3.9.4:', type='build', when='@1.3')
+    depends_on('cmake@3.10:', type='build', when='@1.4.0:1.4.7')
+    depends_on('cmake@3.11:', type='build', when='@1.4.8:')
+    depends_on('cmake@3.12:', type='build', when='@develop')
```
ChristianTackeGSI: Please remember to add a new line here for the next fairmq release, that depends on cmake 3.12.
Or could we just assume, that the next release is `1.4.13`? Then `when='@1.4.13:'` would be better?

packages/fairmq/package.py
dennisklein: done

packages/fairmq/package.py
dennisklein: done

packages/fairmq/package.py
dennisklein: This is definitely something the package maintainer has to get right. While it is likely the next release is indeed `v1.4.13`, IMHO the package should not try to guess anything.

dennisklein:
@kresan I have included the two commits from https://github.com/kresan/FairSoft/tree/test_feb20 in this PR so far. I squashed them together and also did some additional renaming of files as discussed with @ChristianTackeGSI today. FYI, the structure is now as follows:

```
env
├── dev
│   ├── sim_no-threads.yaml
│   ├── sim_threads_no-x_no-opengl.yaml
│   └── sim_threads.yaml
└── jun19
    ├── sim_no-threads.yaml
    └── sim_threads.yaml
test/env
├── dds.yaml
├── fairlogger.yaml
└── fairmq.yaml
```

* `/env` shall contain the spack environments intended to be consumed by FairSoft users. Your changes in the `feb20` directory are now in `/env/dev`. The idea for the workflow here is to mutate dev environments (now part of the CI builds) until they are ready for the release and then copy them to the `feb20` in the release commit or close to the final release commit.
* `test/env` shall contain the spack environments intended purely for testing purposes.

The renaming bit was squashed into your commit, we assumed you will be fine with it. Of course, we can always still change it.

Since this PR is ready to be merged very soon (probably tomorrow), pls rebase your following work regarding the `feb20` already on this branch.

dennisklein:
@kresan I removed the bump of Geant4 from this PR, because it is not yet working. Some more work is needed on the `geant4-data`-family of packages. But this should happen in a different PR.

ChristianTackeGSI:
We are at 22 commits on this currently PR. I find this quite a lot.

And all the CI related things have no relation to the current topic of this PR.
I moved most of the commits about the CI system into their own PR (and branch). See #186.
I hope this makes things easier to follow.

dennisklein:
Let's get #186 merged. Then I will rebase this PR on it and then let's merge it, too. The FairMQ and ODC trees work as far as I can see. The work on Geant4 and Root can follow in separate PRs. Otherwise, this PR will really blow up and never end ;) I think it is fine if the development environments are broken for a few days in the `dev` branch.

env/dev/sim_no-threads.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - when: platform == 'linux'
+      root: [root@6.18.04+fortran+gdml+http+pythia6+pythia8+vc~vdt~python~tmva]
+    - when: platform == 'darwin'
+      root: [root@6.18.04+fortran+gdml+http+pythia6+pythia8+vc~vdt~python~tmva+aqua]
+  specs:
+    - pcre+jit
+    - pythia6@428-alice1
+    - hepmc@2.06.09 length=CM momentum=GEV
+    - pythia8@8301
+    - geant4@10.05.p01~qt~vecgeom~opengl~x11~motif~data~clhep+threads
```
ChristianTackeGSI: Just noticed:
The env is called `no-threads`, but here we have `+threads`.

packages/fairmq/package.py
```diff
@@ -9,74 +12,77 @@ class Fairmq(CMakePackage):
...
+            values=('11', '14', '17'),
+            multi=False,
+            description='Force the specified C++ standard when building.')
+    conflicts('cxxstd=11', when='@1.4.11:')
+
+    patch('fix_find_dds.patch', when='@1.4.0:1.4.4')
+
+    depends_on('googletest@1.7:', when='@:1.4.8')
+    depends_on('boost@1.64: +container+program_options+thread+system+filesystem+regex+date_time', when='@1.3')
+    depends_on('boost@1.64: +container+program_options+filesystem+date_time+regex', when='@1.4')
```
ChristianTackeGSI: Where is the boost dependency for `@develop`?
I start to wonder, why the unit test works for `@develop`.
Looking at the last cdash output, boost is missing from the `fairmq@develop` concretization… strange.

Fix is easy: turn `1.4` into `1.4:`. But I'd like to understand, why the unit test works.

packages/fairmq/package.py
```diff
@@ -9,74 +12,77 @@ class Fairmq(CMakePackage):
...
+    conflicts('^boost@1.70:', when='^cmake@:3.14')
+    depends_on('fairlogger@1.2:1.5', when='@:1.4.7')
+    depends_on('fairlogger@1.2:', when='@1.4.8:,develop')
```
ChristianTackeGSI: `@1.4.8:` should already include `@develop`.

env/dev/sim_no-threads.yaml
dennisklein: good catch, thats a copy&paste fail when I reverted the Geant4 bump, will correct

packages/fairmq/package.py
dennisklein: > But I'd like to understand, why the unit test works.

By default, the CMake command `find_package` is trying to be very smart about finding a suitable dependency, see https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure. My suspicion is that even though spack does not provide an explicit hint, CMake is still able to find a suitable Boost. I will investigate and add the necessary flags to switch off this behaviour.

packages/fairmq/package.py
dennisklein: true, will remove it

packages/fairmq/package.py
dennisklein: Turns out the explanation is even simpler: Even though Boost is not concretized as a direct dependency of FairMQ, spack still puts it into the `$CMAKE_INSTALL_PREFIX`. My guess is that the `setup_dependent_build_environment` of transitive dependencies is not filtered out (because the spack dependency model has no knowledge about which of them are public dependencies - they have to leave them in).

PR #184, opened 2020-02-04
FIX: cmake inside ctest failing

ChristianTackeGSI:
We currently use ctest to orchestrate our tests.
The tests then call spack.
Which then calls cmake.

And it happens to us, that the called cmake isn't the one
built by spack, but the one from the system.

So here is a test case that checks this.
Yes, this will fail on our CI system.  But we need a good
simple test case.

To call on your desktop:

* comment the submit part in FairSoft_test.cmake
* comment the "rm -rf" in test/buildclean.sh
* Use "spack -d install -v" in test/buildsimple.sh
* Call ctest -VV -S FairSoft_test.cmake -R cmake

### DIFF
--- a/packages/cmake/package.py
+++ b/packages/cmake/package.py
@@ -128,7 +128,13 @@ class Cmake(Package):
     conflicts('%intel@:14', when='@3.14:',
               msg="Intel 14 has immature C++11 support")
 
+    sanity_check_is_file = ['bin/cmake']
+
     phases = ['bootstrap', 'build', 'install']
+
+    def setup_build_environment(self, env):
+        super(Cmake, self).setup_build_environment(env)
+        env.unset('DASHBOARD_TEST_FROM_CTEST')
 
     def flag_handler(self, name, flags):
         if name == 'cxxflags' and self.compiler.name == 'fj':
@@ -197,3 +203,5 @@ class Cmake(Package):
 
     def install(self, spec, prefix):
         make('install')
+
+        self.sanity_check_prefix()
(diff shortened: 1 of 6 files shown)


ChristianTackeGSI:
Let's wait for the build, then please review/merge.

packages/cmake/package.py
```diff
@@ -0,0 +1,207 @@
...
+        if '+doc' in spec:
+            args.append('--sphinx-html')
+            args.append('--sphinx-man')
+
+        # Now for CMake arguments to pass after the initial bootstrap
+        args.append('--')
+
+        # Make sure to create an optimized release build
+        args.append('-DCMAKE_BUILD_TYPE=Release')
+
```
dennisklein:
```diff
+ # Install CMake correctly, even if the `spack install` runs inside a ctest environment
+ args.append('-DCMake_TEST_INSTALL=OFF')
+
```

This also works (I tested it) instead of unsetting the `$DASHBOARD_TEST_FROM_CTEST` (do we know what other effects this may have?). See https://gitlab.kitware.com/cmake/cmake/blob/master/Tests/CMakeInstall.cmake#L1-6.

ChristianTackeGSI: 1. I think, unsetting `$DASHBOARD_TEST_FROM_CTEST` shouldn't hurt. As in "get the build environment as reproducible as possible".
2. Adding the `-DCMake_TEST_INSTALL=OFF` looks also good.

What about doing both?

dennisklein: All fine with me, I will also merge it how it is now. I leave it to your decision.

ChristianTackeGSI:
Okay, I'd say: Merge as is.

I'll try to use the `-D` thing when upstreaming this.

PR #190, opened 2020-02-14
DDS: Fix wn_bin target

dennisklein:
@kresan I believe this fixes the `wn_bin` topic for good, not even the modification of the `compiler.yaml` file is needed any more. But someone needs to talk to the DDS team if there is any chance they will eventually accept these changes upstream, maybe we can discuss it on Monday in the meeting?

For reference, this branch contains the patch against DDS `master`: https://github.com/dennisklein/DDS/tree/modernize_wn_bin. It is missing a small part that touches a case when they build some additional test targets which we do not in FairSoft - if they are open to upstreaming this, I will of course add it.


packages/dds/fix_wn_bin_2.2.patch
```diff
@@ -0,0 +1,185 @@
...
++list(JOIN EXCLUDED_SYSTEM_PREFIXES "|" excluded_system_prefix_regex)
++set(excluded_system_prefix_regex "^(${excluded_system_prefix_regex})/")
++
++# Copy dependencies to wn pkg staging directory
++foreach(file IN LISTS EXECUTABLES deps)
++  if(NOT file MATCHES ${excluded_system_prefix_regex})
++    message("WN PKG prerequisite='${file}'")
++    file(COPY ${file} DESTINATION ${DESTINATION})
++  endif()
++endforeach()
```
ChristianTackeGSI: Could you consider moving this into a `fix_wn_bin_common.patch` to reduce duplication?

dennisklein: Yes, I will look into deduplicating the patches.

dennisklein: So I looked over the diffs between the various patches and the part you have selected above is really the only common part which could be deduplicated. But I do not think it is meaningful to have it as a separate patch file without the calling code - it will just add overhead to generating the patch file (Right now it is straightforward, just rebase https://github.com/FairRootGroup/DDS/pull/292/commits/afe07196b3d4ccc0db580dcfe9425f46e014a40a and https://github.com/FairRootGroup/DDS/pull/292/commits/3d0aad8ff10e8ffa9cf129df0d1a5b99878cd7ed onto any given version and run e.g. `git diff -p upstream/master HEAD > ../FairSoft/packages/dds/fix_wn_bin_master.patch`) and saves a few bytes. Or, which concerns do you have if those lines are duplicated other than it costing a few bytes?

dennisklein: (Btw, I made a mistake in the master patch file, it is a bit shorter now)

ChristianTackeGSI: It was mostly a matter of DRY.
And maybe a matter of maintainance. If we ever change the common part, we would only need to fix it once for all of them.

That said, we most likely wont have to maintain these patches forever, so I am happy with whatever works.

kresan:
If fixing wn_bin requires cmake upgrade in jun19 - I am fine with that.

ChristianTackeGSI:
@dennisklein could you rebase and fix the conflicts?

dennisklein:
Rebased on current `dev`.

PR #192, opened 2020-02-19
WIP: Sync our recipes somewhat with upstream

ChristianTackeGSI:
Target:
* reduce diff to upstream spack
* get newer versions
* get updated build structure
* finally drop our forked recipes

Currently:
* New versions for hepmc
* Some recipe updates for root
* Some test to get root 6.18 compile


ChristianTackeGSI:
I will do that in other PRs. This doesn't fit very well any more.

PR #193, opened 2020-02-20
Add python2 build dependency for root@6.16

kresan:
Build mesa without llvm on linux
Build python@2.7.16 as build dependency for root.
Fix both in jun19 environments.
In root recipe, if @6.16 then pass PYTHON_EXECUTABLE

### DIFF
--- a/env/jun19/sim_no-threads.yaml
+++ b/env/jun19/sim_no-threads.yaml
@@ -1,11 +1,12 @@
 spack:
   definitions:
     - when: platform == 'linux'
-      root: [root@6.16.00+fortran+gdml+http+memstat+pythia6+pythia8+vc~vdt~python~tmva]
+      root: [root@6.16.00+fortran+gdml+http+memstat+pythia6+pythia8+vc~vdt~python~tmva ^mesa~llvm]
     - when: platform == 'darwin'
       root: [root@6.16.00+fortran+gdml+http+memstat+pythia6+pythia8+vc~vdt~python~tmva+aqua]
   specs:
     - pcre+jit
+    - python@2.7.16
     - cmake@3.13.4
     - googletest@1.8.1
     - boost@1.68.0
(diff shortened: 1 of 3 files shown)


env/jun19/sim_no-threads.yaml
```diff
@@ -1,7 +1,7 @@
 spack:
   definitions:
     - when: platform == 'linux'
-      root: [root@6.16.00+fortran+gdml+http+memstat+pythia6+pythia8+vc~vdt~python~tmva]
+      root: [root@6.16.00+fortran+gdml+http+memstat+pythia6+pythia8+vc~vdt~python~tmva ^mesa~llvm ^python@2.7.16]
```
ChristianTackeGSI: Could you try moving the `^python@2.7.16` as a `- python@2.7:2.99` as its own entry in the `specs:` section to the `prce+jit` part to aid the concretizer?

packages/root/package.py
```diff
@@ -505,6 +506,10 @@ def cmake_args(self):
             options.append('-DCMAKE_PROGRAM_PATH={0}'.format(
                 self.spec['mysql-client'].prefix.bin))
 
+        if '@6.16' in self.spec:
+            options.append('-DPYTHON_EXECUTABLE={0}/python'.format(
+                self.spec['python'].prefix.bin))
+
         if '+python' in self.spec:
```
ChristianTackeGSI: What about this?

```suggestion
        if '+python' in self.spec or '@6.16' in self.spec:
```

env/jun19/sim_no-threads.yaml
ChristianTackeGSI: Okay, I think, you can leave out the `^python@2.7.16` bit completely, because the `depends_on` should take care of it. This works nicely on my local machine.

kresan: Test shows that the concretizer fails, if one does not specify python@2.7.16 in the environment

PR #198, opened 2020-03-03
Bump Geant4, Pythia8, ROOT, VGM.

kresan:
Bump geant4-data
Add vmc package
Update dev environments


packages/flatbuffers/package.py
```diff
@@ -26,7 +26,7 @@ class Flatbuffers(CMakePackage):
     def cmake_args(self):
         args=[]
         args.append('-DFLATBUFFERS_BUILD_SHAREDLIB=ON')
-        args.append('-DFLATBUFFERS_BUILD_FLATLIB=OFF')
+        args.append('-DFLATBUFFERS_BUILD_FLATLIB=ON')
```
ChristianTackeGSI: Do you actually need static libraries?
They break fairmq often.

packages/vmc/package.py
```diff
@@ -0,0 +1,38 @@
...
+    homepage = "https://github.com/vmc-project/vmc"
+    url      = "https://github.com/vmc-project/vmc/archive/v0-1.tar.gz"
+
+    version('1-0', sha256='3da58518b32db1b503082e3205884802a1a263a915071b96e3fd67861db3ca40')
+
+    depends_on('root')
+
+    def cmake_args(self):
+        options = []
+        return options
```
ChristianTackeGSI: Please drop this, if it's not needed.

```suggestion
```

packages/fairroot/package.py
```diff
@@ -40,6 +40,8 @@ class Fairroot(CMakePackage):
     # Dependencies for dev version
     depends_on('boost@1.68.0: cxxstd=11 +container')
 
+    depends_on('vmc', when='@dev')
+
```
ChristianTackeGSI: Has this changed in fairroot?
Or did the old versions need flatbuffers also, just we didn't specifiy it here?

ChristianTackeGSI: Okay, my fault. No idea why I haven't seen the `@dev`.

packages/flatbuffers/package.py
kresan: ok, reverted back. But we need to make change in FairRoot that it does not search for static library of flatbuffers.

packages/vmc/package.py
kresan: done

kresan:
There was a change in upstream git recipe, which we still do not have. A missing dependency was added. I have put it to our side, till we update Spack.

packages/flatbuffers/package.py
ChristianTackeGSI: Okay, that's a little strange.

@dennisklein can you please comment on the following, I seem to miss something.

* flatbuffers can be built as static and/or shared libraries.
* If building static libraries at all, fairmq (some versions?) tried to link a static library into a shared object. This fails. Why fairmq isn't using the shared flatbuffers: I don't know.
* If building only shared flatbuffers, fairroot seems to have an issue: Why can't it link to the shared library?

What should we do?

dennisklein: The Flatbuffers library is exports two different names for the static and shared library: `flatbuffers::flatbuffers` and `flatbuffers::flatbuffers_shared`. FairMQ initially only used `flatbuffers::flatbuffers` and only some patch release versions later gained the CMake logic to use `flatbuffers::flatbuffers_shared` is the other is not available.

FairRoot currently only relies on the `flatbuffers::flatbuffers` name.

dennisklein: The only requirement is that the static flatbuffers is built with PIC. For versions that do not use the shared name, the relevant lines to patch are:

https://github.com/FairRootGroup/FairRoot/blob/4ab2d2b1f0392ad93b92026d70fe049c38665f7d/examples/advanced/Tutorial3/CMakeLists.txt#L127

dennisklein: Hm, FairMQ should always support both names, the Flatbuffers dependency was introduced in `v1.4.9`.

ChristianTackeGSI: Shouldn't the logic be the other way round?
First check for `flatbuffers::flatbuffers_shared`, and if that does not exist, use `flatbuffers::flatbuffers`? When creating a shared library/object (fairmq! fairoot?), one need the library as a shared library.

dennisklein: Sure, I don't mind switching the logic, if you prefer. But why is it needed?

Btw, usually, most CMake packages export only one name and either populate it with the static or the shared library.

ChristianTackeGSI: Okay, after some private discussions, the current conclusion seems to be:

* Let fairroot use the shared version of flatbuffers. @dennisklein will take care of this.
* As a quick short fix, one could remove the `depends_on('flatbuffers')` in `fiarroot/package.py`.

dennisklein: FairRootGroup/FairRoot#936

kresan:
FairRoot issue with protobuf still persists:
`  >> 137    CMake Error at CMakeLists.txt:517 (get_target_property):
     138      get_target_property() called with non-existent target
     139      "flatbuffers::flatbuffers".
`

ChristianTackeGSI:
About fairroot and flatbuffers:
I think, @dennisklein wanted to take care about it.
See: https://github.com/FairRootGroup/FairRoot/pull/936

packages/fairroot/package.py
```diff
@@ -80,6 +82,8 @@ def cmake_args(self):
         self.spec['boost'].prefix))
         options.append('-DBOOST_LIBRARYDIR={0}/lib'.format(
         self.spec['boost'].prefix))
+        options.append('-DFlatbuffers_DIR={0}'.format(
+        self.spec['flatbuffers'].prefix))
```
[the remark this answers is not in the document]
kresan: I think it will work also without

packages/vmc/package.py
```diff
@@ -0,0 +1,34 @@
...
+
+class Vmc(CMakePackage):
+    # The Virtual Monte Carlo core library.
+
+    homepage = "https://github.com/vmc-project/vmc"
+    url      = "https://github.com/vmc-project/vmc/archive/v0-1.tar.gz"
+
+    version('1-0', sha256='3da58518b32db1b503082e3205884802a1a263a915071b96e3fd67861db3ca40')
+
+    depends_on('root')
```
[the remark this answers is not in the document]
kresan: officially, vmc requires root starting from 6.18.06. but it was also adopted to root 6.18.04. so we can take the second one.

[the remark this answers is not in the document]
ChristianTackeGSI: I'll create a "make things nicer" PR today or tomorrow.
I'll include that information in it.

And I'll try to remove the `-DFlatbuffers_DIR=` thing there.

Issue #194, opened 2020-02-23
[request] /cvmfs/fairroot.gsi.de: add ROOT python bindings


klenze:
Hi,
I would like to request that the future versions of the binary FairSoft installation provided at /cvmfs/fairroot.gsi.de/ on lxpool.gsi.de and friends are configured with python bindings for ROOT.

Cheers,
   Philipp

kresan:
ROOT from Spack-based FairSoft is now compiled with python bindings (#202 ). Corresponding binaries will be later installed on cvmfs

PR #202, opened 2020-03-21
Build root with tmva support

kresan:
PANDA needs root with tmva.

Previously, the idea was to skip tmva in order to avoid python dependency completely. Now we have it anyhow, event pinned down in jun19 environments. So it does not change a lot having tmva feature on again.

### DIFF
--- a/env/dev/sim_no-threads.yaml
+++ b/env/dev/sim_no-threads.yaml
@@ -1,9 +1,9 @@
 spack:
   definitions:
     - when: platform == 'linux'
-      root: [root@6.18.04+fortran+gdml+http+pythia6+pythia8+vc~vdt~python~tmva]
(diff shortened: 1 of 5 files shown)


ChristianTackeGSI:
* For jun19, I think, you should add a `py-numpy@1.16.5`, because 1.16 seems to be the last version supporting python2.
* In general, could you please use `+tmva`? The root package in spack upstream moved to `~tmva` being the default.
* Maybe use `+python` too? If tmva pulls it in, we can enable that part anyway? #194 comes to my mind.

kresan:
@ChristianTackeGSI have implemented all your suggestions. let's wait for tests

kresan:
@ChristianTackeGSI tests are successful now (for jun19)

PR #205, opened 2020-03-26
Import intel-tbb package

kresan:
Change sha256 for intel-tbb@2019.4
Add fairsoft.intel-tbb as ebb provider

With these changes "-C ./config" is mandatory also for linux

### DIFF
--- a/config/packages.yaml
+++ b/config/packages.yaml
@@ -1,0 +1,5 @@
+packages:
+  all:
+    providers:
+      tbb: [fairsoft.intel-tbb]
+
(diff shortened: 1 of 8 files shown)


config/packages.yaml
```diff
@@ -0,0 +1,5 @@
+packages:
+  all:
+    providers:
+      tbb: [fairsoft.intel-tbb]
```
ChristianTackeGSI: Wow, I hadn't exptected that to be needed.

PR #208, opened 2020-03-27
fairroot: Support building against shared flatbuffer lib

dennisklein:
Fix on package level for now to unblock the work towards upcoming releases. Maybe I will find the time to fix it properly upstream and separate from FairRootGroup/FairRoot#933 before the FairSoft release, then this can go away again.


dennisklein:
> Let's wait for the builds?

Sure.

PR #209, opened 2020-03-30
Create recipe for r3broot

ChristianTackeGSI:
* r3broot needs patching to find fairlogger
* Created the basic recipe for it
* Create an environment for it, basicly copied from jun19
* Put environment in unit tests

Notes:
* The ctest ("make test") has been disabled, because it does not work fully.

Also:
* Changed fairroot to not any more output utf8 characters


CMakeLists.txt
```diff
@@ -73,6 +73,8 @@ fs_test_env(NAME dev.threads_nox_noopengl ENVFILE env/dev/sim_threads_no-x_no-op
 fs_test_env(NAME jun19.threads   ENVFILE env/jun19/sim_threads.yaml)
 fs_test_env(NAME jun19.nothreads ENVFILE env/jun19/sim_no-threads.yaml)
 
+fs_test_env(NAME r3broot ENVFILE env/r3broot/dev.yaml)
```
dennisklein: This is introducing a new naming scheme that is conflicting with the one we have at the moment which is `env/<version>/<variant>.yaml`. What about `env/dev/r3broot.yaml`?

kresan:
All this needs to go to experiment repo. This will be not our responsibility. Experiments will have their own recipes and they will maintain them. Please send this PR to [https://github.com/R3BRootGroup/FairSoft](https://github.com/R3BRootGroup/FairSoft)

CMakeLists.txt
kresan: agree with Dennis

env/r3broot/dev.yaml
```diff
@@ -0,0 +1,27 @@
...
+      root: [root@6.16.00+fortran+gdml+http+memstat+pythia6+pythia8+vc~vdt+python+tmva ^mesa~llvm]
+    - when: platform == 'darwin'
+      root: [root@6.16.00+fortran+gdml+http+memstat+pythia6+pythia8+vc~vdt+python+tmva+aqua]
+  specs:
+    - pcre+jit
+    - python@2.7.16
+    - py-numpy@1.16.5
+    - googletest@1.8.1
+    - boost@1.68.0
+    - fairlogger
```
kresan: Concretizer fails because version of FairLogger is not specified

CMakeLists.txt
ChristianTackeGSI: Renamed.

env/r3broot/dev.yaml
ChristianTackeGSI: Interestingly enough, the concretizer only fails on mac. I have no idea, what's happening there. Anayway, added the version number from jun19.

env/dev/r3broot.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - py-numpy@1.16.5
+    - googletest@1.8.1
+    - boost@1.68.0
+    - fairlogger@1.4.0
+    - dds@2.4
+    - fairmq@1.4.3
+    - pythia6@428-alice1
+    - hepmc@2.06.09 length=CM momentum=GEV
+    - pythia8@8240
+    - geant4@10.05.p01~qt~vecgeom~opengl~x11~motif~data~clhep+threads
```
kresan: Maybe this helps for crashing r3broot simulations. R3BRoot can be used only with geant4 single-threaded. Please put ~threads

env/r3broot/dev.yaml
kresan: Was also failing in my Debian 8 container. Now is fine.

CMakeLists.txt
ChristianTackeGSI: marking resolved. If that's a problem, please complain.

env/dev/r3broot.yaml
ChristianTackeGSI: done.
I also put a `depends_on('geant4~threads')` in r3broot, because it's a strict dependency of r3broot it seems.

marking resolved, if not, please complain/unmark.

packages/r3broot/package.py
```diff
@@ -0,0 +1,40 @@
...
+    depends_on('geant3')
+    depends_on('geant4~threads')
+    depends_on('geant4_vmc')
+    depends_on('pythia6')
+    depends_on('pythia8')
+
+    patch('fairlogger_include.patch')
+
+    def build(self, spec, prefix):
+        super(R3broot, self).build(spec, prefix)
```
kresan, line 33: on Debian 8 with gcc 4.9.2 cxx flag `-std=c++11` is missing. compilation fails

dennisklein: `-DCMAKE_CXX_STANDARD=11` will add it

ChristianTackeGSI: Added `cmake_args()`. Let's see.

PR #210, opened 2020-04-02
root: Use some more spack packages

ChristianTackeGSI:
root can build a bunch of packages internally. Many of them are available via spack. By using the spack ones, those parts can be build once for all versions/variants of root.

So use the spack packages instead of the internal ones:
* freetype
* fftw

### DIFF
--- a/repos/fairsoft/packages/root/package.py
+++ b/repos/fairsoft/packages/root/package.py
@@ -15,6 +15,8 @@ class Root(CMakePackage):
     homepage = "https://root.cern.ch"
     url      = "https://root.cern/download/root_v6.16.00.source.tar.gz"
 
+    maintainers = ['chissg', 'HadrienG2', 'drbenmorgan']
+
     # ###################### Versions ##########################
 
     # Master branch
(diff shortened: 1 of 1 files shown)


ChristianTackeGSI:
@fuhlig1
I requested you for review, because you once changed some of those packages to "use root's internal ones". As all seems to build fine, I can't say, what was wrong when you decided to switch it.

kresan:
I would wait with changing the way we build root until release is done. Then we reproduce the jun19 behaviour and can change it in the dev. We can not underestimate the fact, that experiments should be able to cross check the results with old / new FairSoft.

fuhlig1:
@ChristianTackeGSI 

I am not sure why I did this at the time but I think I wanted to reproduce the way we did the build before. Maybe there was also a problem with the packages but I can't recall. I also would prefer to wait till the release is done and do the change afterwards.

ChristianTackeGSI:
As we decided to have the release in August/September, we should have enough time to reconsider this PR.

kresan (approved):

Issue #213, opened 2020-04-07
Geant4 bug in version 10.4 and newer


fuhlig1:
Recently a bug was found in  GEANT4 versions 10.4 and newer.
GEANT4 developers proposed the following quick fix:

https://bugzilla-geant4.kek.jp/show_bug.cgi?id=2226#c8

Some tests with CBMROOT were made (we have GEANT4 10.5.1 in the latest FairSoft release) 
which confirmed that CBM is affected by the bug. (see https://redmine.cbm.gsi.de/issues/1647)

A fix was applied on the CBMROOT level when initializing Geant4 but it would be much better to fix the issue in FairSoft if the new version of Geant4 is still affected.

fuhlig1:
I rechecked the Geant4 bug report and there it is clearly stated that the problem will be fixed in versions > 10.6, so we are affected by the bug. Since the bugfix is a one line correction in one of the source files I would propose to add a patch to our recipe and fix the bug as described in the bug report.

ChristianTackeGSI:
The patch has been merged. If you have a chance, please test and confirm, that this fixes your issues.

PR #214, opened 2020-04-07
geant4: Bug Workaround for Isomer Transport

ChristianTackeGSI:
Patching @10.04:10.06 with the hint in
https://bugzilla-geant4.kek.jp/show_bug.cgi?id=2226#c7

fixes: #213 
cc: @fuhlig1


ChristianTackeGSI:
> If the function filter_file exchange the first line by the second […]

With `string=True` it replaces all ocurences of the first string by the second string.

Let's wait for the build to finish. I don't expect surprises, but anyway.

PR #215, opened 2020-04-07
Limit number of executors for those long running jobs

dennisklein:
Trying the https://plugins.jenkins.io/throttle-concurrents/ plugin. Will start the CI later manually to learn if this does what I think it does.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -68,7 +68,9 @@ pipeline {
               ${ctestcmd}
             """
           }
-          parallel(linux_jobs + macos_jobs)
+          throttle(['long']) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -68,7 +68,7 @@ pipeline {
               ${ctestcmd}
             """
           }
-          parallel(linux_jobs + macos_jobs)
+          throttle(['long']) { parallel(linux_jobs + macos_jobs) }
```
ChristianTackeGSI, line 71: Could you please move the contents of the braces on its own line? So that we have one "building block" per line?

dennisklein: sure

PR #217, opened 2020-04-07
Allow enabling spack's ccache usage

dennisklein:


dennisklein:
> Looks good. I assume, this PR itself will already utilize the ccache on one machine?

Yes, on `witmac007` I set the environment variable accordingly in the jenkins agent settings. There `ccache` is installed and configured for the CI user.

PR #218, opened 2020-04-07
CI: Point ccache to the shared cache location

dennisklein:
Because `$HOME` is changed, `ccache` does not find the location any more by
default.

### DIFF
--- a/test/buildsetup.sh
+++ b/test/buildsetup.sh
@@ -26,6 +26,12 @@ else
 	echo "*** Reusing test directory: $TARGETDIR"
 fi
 
+if [ "$SPACK_CCACHE" = "true" ]
+then
+  CCACHE_CACHEDIR="$(ccache -k cache_dir)"
+  CCACHE_MAXSIZE="$(ccache -k max_size)"
+fi
+
 export HOME="$TARGETDIR"
 
 if [ ! -d "$HOME/.spack/" ]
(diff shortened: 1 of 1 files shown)


test/buildsetup.sh
```diff
@@ -26,6 +26,12 @@ else
 	echo "*** Reusing test directory: $TARGETDIR"
 fi
 
+if [ "$SPACK_CCACHE" = "true" ]
+then
+  CCACHE_CACHEDIR="$(ccache -k cache_dir)"
+  CCACHE_MAXSIZE="$(ccache -k max_size)"
+fi
+
```
ChristianTackeGSI, line 34: This looks like you're transferring specific options. What if another option is being added later on?

What about symlinking the appropiate directories into the new HOME?

dennisklein: > What about symlinking the appropiate directories into the new HOME?

I thought about symlinking the config file, but then I would have to write the cache dir explicitely into the config file and one would have to know to not remove it even though it is the default path. Hmm, I could copy the config file and write the cache dir explicitely..

ChristianTackeGSI: Well, I meant symlink the config file and symlink the cache directory (it's probably relative to the home?)

But anyway: If this works, merge it.

dennisklein:
The effect is a ~20% speedup between two consecutive builds of this PR.

test/buildsetup.sh
dennisklein: ya, let's refine, when it breaks.

PR #221, opened 2020-04-09
boost: Fix for Version Comparison on Newer Clang on Darwin

ChristianTackeGSI:
Applying to boost @1.56.0:1.72.0 as that's the version range, where this patch can be cleanly applied. I was not able to test this.

See: https://github.com/boostorg/build/issues/440
See: https://github.com/macports/macports-ports/pull/6726

### DIFF
--- a/packages/boost/darwin_clang_version.patch
+++ b/packages/boost/darwin_clang_version.patch
@@ -1,0 +1,31 @@
+From 40960b23338da0a359d6aa83585ace09ad8804d2 Mon Sep 17 00:00:00 2001
+From: Bo Anderson <mail@boanderson.me>
+Date: Sun, 29 Mar 2020 14:55:08 +0100
+Subject: [PATCH] Fix compiler version check on macOS
+
+Fixes #440.
+---
+ src/tools/darwin.jam | 5 +++--
+ 1 file changed, 3 insertions(+), 2 deletions(-)
+
+diff --git a/src/tools/darwin.jam b/src/tools/darwin.jam
+index 8d477410b0..97e7ecb851 100644
+--- tools/build/src/tools/darwin.jam
++++ tools/build/src/tools/darwin.jam
+@@ -137,13 +137,14 @@ rule init ( version ? : command * : options * : requirement * )
+     # - Set the toolset generic common options.
+     common.handle-options darwin : $(condition) : $(command) : $(options) ;
+     
++    real-version = [ regex.split $(real-version) \\. ] ;
+     # - GCC 4.0 and higher in Darwin does not have -fcoalesce-templates.
+-    if $(real-version) < "4.0.0"
++    if [ version.version-less $(real-version) : 4 0 ]
+     {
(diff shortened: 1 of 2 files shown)


kresan:
I will test it now on my Mac

packages/boost/darwin_clang_version.patch
```diff
@@ -0,0 +1,31 @@
...
+From: Bo Anderson <mail@boanderson.me>
+Date: Sun, 29 Mar 2020 14:55:08 +0100
+Subject: [PATCH] Fix compiler version check on macOS
+
+Fixes #440.
+---
+ src/tools/darwin.jam | 5 +++--
+ 1 file changed, 3 insertions(+), 2 deletions(-)
+
+diff --git a/src/tools/darwin.jam b/src/tools/darwin.jam
```
kresan, line 11:
```suggestion
 tools/build/src/tools/darwin.jam | 5 +++--
   1 file changed, 3 insertions(+), 2 deletions(-)
 
  diff --git a/tools/build/src/tools/darwin.jam b/tools/build/src/tools/darwin.jam
```

packages/boost/darwin_clang_version.patch
```diff
@@ -0,0 +1,31 @@
+From 40960b23338da0a359d6aa83585ace09ad8804d2 Mon Sep 17 00:00:00 2001
+From: Bo Anderson <mail@boanderson.me>
+Date: Sun, 29 Mar 2020 14:55:08 +0100
+Subject: [PATCH] Fix compiler version check on macOS
+
+Fixes #440.
+---
+ src/tools/darwin.jam | 5 +++--
+ 1 file changed, 3 insertions(+), 2 deletions(-)
+
```
kresan, line 10: Can we skip this?

packages/boost/darwin_clang_version.patch
```diff
@@ -0,0 +1,31 @@
...
+diff --git a/src/tools/darwin.jam b/src/tools/darwin.jam
+index 8d477410b0..97e7ecb851 100644
+--- tools/build/src/tools/darwin.jam
++++ tools/build/src/tools/darwin.jam
```
kresan, line 14:
```suggestion
a/tools/build/src/tools/darwin.jam
b/tools/build/src/tools/darwin.jam
```

kresan:
With proposed changes boost is successfully patched and compiled

packages/boost/darwin_clang_version.patch
ChristianTackeGSI: *GRR* I missed the `level=0` in the `patch` command. Fixing.

ChristianTackeGSI:
Okay, updated `package.py`. Can you retry, please?

packages/boost/darwin_clang_version.patch
ChristianTackeGSI: Do we really need to?
My test on a mac showed no problems after I added the `level=0`.

kresan: with level=0 it works

PR #227, opened 2020-04-22
vmc: new Version, Improve Dictionary Generation

ChristianTackeGSI:
### Add version 1-0-p1

This version adds a fix for root dictionaries.
This improves using vmc in the root interpreter.

Problem with 1-0:
```console
root [0] TMCManager *tmc;
Error in <TInterpreter::TCling::AutoLoad>: failure loading library libVM
```

With 1-0-p1 there are now problems with headers: When the source/build tree exists, everything works fine. But if it is removed, then headers are not found, even with `ROOT_INCLUDE_PATH` set.

### Make Header Paths in Dictionary Relative

According to

https://root.cern/doc/v618/release-notes.html#header-location-and-root_generate_dictionary-root_standard_library_package

the listed headers in ROOT_GENERATE_DICTIONARY should use relative paths for the listed headers.

With this change, the headers in the install tree are found by the interpreter, when `ROOT_INCLUDE_PATH` is set.

### Let Headers be Searched in the Install Tree

When loading a module in the root interpreter, it needs to load some more headers.
Those headers are searched for in $ROOT_INCLUDE_PATH and a list of paths inside the shared library module.

The list of paths gets all the entries from the include search path (`-I` arguments) and then is reduced by the `-excludePath` options to `rootcling`.

So the idea is:
* Add the install tree path to the _end_ of the `-I` list, so that during generation the source tree headers are being used.
* Use `-excludePath` to remove nearly all source/build paths from the list that is being stored in the shared library. This way, there is no security problem with searching in the source/build tree.

### DIFF
--- a/env/dev/sim_no-threads.yaml
+++ b/env/dev/sim_no-threads.yaml
@@ -11,7 +11,7 @@ spack:
     - pythia8@8301
     - geant4@10.05.p01~qt~vecgeom~opengl~x11~motif~data~clhep~threads
     - $root
-    - vmc@1-0
+    - vmc@1-0-p1
     - geant3@v3-0_fairsoft
     - vgm@4-7
     - geant4_vmc@5-0
--- a/env/dev/sim_threads.yaml
+++ b/env/dev/sim_threads.yaml
@@ -11,7 +11,7 @@ spack:
     - pythia8@8301
     - geant4@10.05.p01~qt~vecgeom~opengl~x11~motif~data~clhep+threads
     - $root
-    - vmc@1-0
+    - vmc@1-0-p1
     - geant3@v3-0_fairsoft
     - vgm@4-7
     - geant4_vmc@5-0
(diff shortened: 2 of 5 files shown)


ChristianTackeGSI:
I have started by upstreaming the first part of the patch (second commit in this series) as https://github.com/vmc-project/vmc/pull/8

The third commit (adding install path to list and trimming the source/build paths) is a bit more tricky: For spack we know, that after the install there will be no source tree. But others might have other use cases.
I could upstream the part that adds the install tree path to the end of the search list. This one shouldn't break anybody's use cases (at least not break anything, that wasn't broken before).

Other points:
* Before merging this PR  I would like to add `LD_LIBRARY_PATH` setting to the spack recipe.
* Also note, that this PR does not change the to-be-released environments yet. Let me know, if I should do that too.

kresan:
I think it is a good idea to have this patch in the upcoming release too

PR #229, opened 2020-04-30
Get the FairRoot 18.4.0 Release going

ChristianTackeGSI:
Major points:
* Add the RC_v18.4.0 branch as `version('18.4.0')`
* Add test environment for FairRoot 18.4 on jun19

Also some small CI bits.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,21 +66,21 @@ function(fs_test_env)
   fs_test_props(env.${ARGS_NAME})
 endfunction()
 
-fs_test_simple(NAME openssl SPEC openssl)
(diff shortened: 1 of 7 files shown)


dennisklein:
Question: Is it possible to factor out a common env.yaml which does not contain the fairrot spec and the include it in the specialized envs?

ChristianTackeGSI:
> Question: Is it possible to factor out a common env.yaml which does not contain the fairroot spec and then include it in the specialized envs?

Doesn't look like this is possible currently. I am still investigating (asking upstream).
And yes, I was looking for something like this when writing the R3BRoot env.

In the worst case, we might create a cmake-or-whatever tool to create a `spack.yaml` from pieces.

PR #258, opened 2020-05-27
DONT-MERGE: Try font-util on mac

ChristianTackeGSI:


ChristianTackeGSI:
I have no idea, why font-util sometimes fails on mac, but does never fail when run standalone.

So I am closing this. We need some other way to analyse this.

PR #263, opened 2020-06-05
Add openmpi 3 environment

kresan:
Use native pmix. Disable pmix dstore. Add rdma-core dependency.
Change with-pmi to with-pmix. Add libevent dependency.
Add perl-datemanip. Add mpi benchmark tests.


ChristianTackeGSI:
For each package, that has a relation to upstream spack, could you please update https://github.com/FairRootGroup/FairSoft/projects/1 ?

If the package needed modifications:
* Add new card to the ["To upstream"](https://github.com/FairRootGroup/FairSoft/projects/1#column-8819194) column
* Please include info, so that anyone working on that knows, what are the exact changes that need to be upstreamed.
* You could also put distinct commits in our tree that explain the exact modifications

If you copied the recipe from a newer version of spack:
* Please add a card to the ["Upstreamed but not tracked in FairSoft"](https://github.com/FairRootGroup/FairSoft/projects/1#column-8819214) column. So that we can check when upgrading spack, which package recipes can be dropped.

packages/osu-micro-benchmarks/package.py
```diff
@@ -0,0 +1,22 @@
...
+
+    depends_on('openmpi')
+
+    def configure_args(self):
+        args = []
+        args.append('CC={0}/bin/mpicc'.format(self.spec['openmpi'].prefix))
+        args.append('CXX={0}/bin/mpicxx'.format(self.spec['openmpi'].prefix))
+        return args
+
+
```
ChristianTackeGSI, line 22: What's wrong with the osu-micro-benchmarks in spack upstream? Even our old version of spack already had this?

kresan: switched to osu-micro-benchmarks from upstream

kresan:
@ChristianTackeGSI I have added entries for openmpi and perl-datemanip to the "to upstream" column

PR #267, opened 2020-06-29
pythia8: Add Patch and new Versions

ChristianTackeGSI:
* Add patch to avoid `sqrt` of a negative numbers in `Ropewalk.cc`
  See: https://github.com/alisw/alidist/pull/2333
  See: https://github.com/alisw/pythia8/commit/a854fb5c250fe7b7b17e4e43f7dcb03e63ee1364
* Fix typo in patch
  Asked in the alisw PR: https://github.com/alisw/alidist/pull/2333#issuecomment-651054503
* Add version 8244 and 8302
  (none are yet used in any environment)

cc: @fuhlig1

### DIFF
--- a/env/dev/sim_no-threads.yaml
+++ b/env/dev/sim_no-threads.yaml
@@ -7,7 +7,7 @@ spack:
   specs:
     - pythia6@428-alice1
     - hepmc@2.06.09 length=CM momentum=GEV
-    - pythia8@8301
+    - pythia8@8302
     - geant4@10.05.p01~qt~vecgeom~opengl~x11~motif~data~clhep~threads
     - $root
     - vmc@1-0-p1
--- a/env/dev/sim_threads.yaml
+++ b/env/dev/sim_threads.yaml
@@ -7,7 +7,7 @@ spack:
   specs:
     - pythia6@428-alice1
     - hepmc@2.06.09 length=CM momentum=GEV
-    - pythia8@8301
+    - pythia8@8302
     - geant4@10.05.p01~qt~vecgeom~opengl~x11~motif~data~clhep+threads
     - $root
     - vmc@1-0-p1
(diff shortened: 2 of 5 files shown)


ChristianTackeGSI:
JFI
 
According to https://github.com/alisw/alidist/pull/2336#issuecomment-651123548
> […] the issue shows only in a very specialised use of Pythia8 features.

I don't know, what that specialised use is.

So I don't know, whether we need to backport this to FairSoft legacy.

fuhlig1:
Hi Christian,

since we never phased the problem I would not back port the patch. If someone reports a problem which we can relate to the issue he has to use the new FairSoft version.

Ciao

Florian

fuhlig1:
faced. This dammed autocorrection.

> Am 30.06.2020 um 08:14 schrieb Florian Uhlig <f.uhlig@gsi.de>:
> 
> phased

preghenella:
specialised use means when enabling the colour ropes model.
The floating point exception is likely to happen in Pythia8.3 series as well, if that part of the code is not protected.
If you do not use the colour ropes model you could safely leave this patch out.

Pythia authors informed us that they will release an official Pythia8.2 version with the bug fixed.
For the ALICE case, we are patching because the issue hits us in this very moment.

ChristianTackeGSI:
Thanks Roberto for the explanation and notes!

So I will wait a bit for the upstream releases and include those versions in this PR as well.

ChristianTackeGSI:
Okay, rebased to current `dev` branch.
Also added some small bits.

@kresan
Should we consider upgrading our environments to the latest version?

kresan:
> @kresan
> Should we consider upgrading our environments to the latest version?

Yes, we need to upgrade the environments for the upcoming release. Except jun19.

ChristianTackeGSI:
Rebased, and upgrade to 8302 in the dev environments.

PR #278, opened 2020-07-15
New package fairsoft-config

### LINKED ISSUE #271: Add fairsoft-config package
which provides a `fairsoft-config` script and add it to the existing environments

### LINKED ISSUE #295: Geant3 needs "-fallow-invalid-boz" when using gfortran 10.2
On MacOS 10.15.6 and fortran installed by brew (Homebrew GCC 10.2.0) I get the error ()

```
/Users/turany/development/FairSoft/legacy/transport/geant3/minicern/lnxgs/ublow.F:22:20:
   22 |       data iblan1/x'20202000'/
      |                    1
Error: Hexadecimal constant at (1) uses nonstandard X instead of Z [see '-fno-allow-invalid-boz']
/Users/turany/development/FairSoft/legacy/transport/geant3/minicern/lnxgs/ublow.F:23:19:
   23 |       data mask1/x'000000ff'/
      |                   1
Error: Hexadecimal constant at (1) uses nonstandard X instead of Z [see '-fno-allow-invalid-boz']
```

This can be solved by adding the flag **-fallow-invalid-boz** to the fortran flags

dennisklein:
Resolves #271
~Resolves #295~
Requires #298 

* [x] Add to environments
* [x] Test coverage
* [x] Docs


ChristianTackeGSI:
I would prefer some distinct PRs for the following parts, which seem look good and should be dealt with independently:

### Provide top-level env script.
It could even do the `spack repo add .`?
You could even do that optionally, after checking
```bash
if [ "$(spack repo list | sed -n -e '/^fairsoft / { s/^[^ ]* *//; p; }')" != "$PWD" ]
then
        spack repo add --scope site .
fi
```

### Transform envs to anonymous envs
Yes, let's go for that.
In the next step, we create some tool, that assembles `spack.yaml` from bits. I already have some ideas for that.

### Transform envs to anonymous envs
Before doing that, we need to make sure, that VAE is still happy. I think, it's not a big issue, as VAE builds from Dima's private repo currently anyway.
And then, I would prefer, that we take a small intermediate / refuelling stop at 0.14.2. That's nearly 100% ready in `dev_new_spack`.

ChristianTackeGSI:
I have cherry-picked the "Provide top-level env script" commit into #281.

So could you remove the commit from here and comment there as needed?

thisfairsoft.sh
```diff
@@ -54,6 +54,8 @@ fairsoft_repo() {
 fairsoft_repo "fairsoft_backports" "fairsoft-backports"
 fairsoft_repo "fairsoft" "fairsoft"
 
+spack clean -ms
+
```
ChristianTackeGSI, line 58: Please also add to `thisvae.sh`.

I did not factor out this, because we're still at two files. But if this becomes a maintainance issue, we should refactor.

env/dev/r3broot/spack.yaml
```diff
@@ -1,26 +1,31 @@
...
-    - pythia8@8240
-    - geant4@10.05.p01~qt~vecgeom~opengl~x11~motif~data~clhep~threads
-    - $root
-    - geant3@2-7_fairsoft
-    - vgm@4-5
-    - geant4-vmc@4-0-p1
-    - fairroot@18.2.1+sim
-    - r3broot
+  - $concretizer_hints
+  - fairsoft-config@jun19
```
ChristianTackeGSI, line 14: As the `r3broot` env is mainly for CI (maybe should be moved to `test/`?), why do you need to add fairsoft-config? Or does the old fairoot-Version (18.2) now really need it?

env/dev/jun19_fairroot_18_4/spack.yaml
```diff
@@ -1,26 +1,30 @@
...
-    - hepmc@2.06.09 length=CM momentum=GEV
-    - pythia8@8240
-    - geant4@10.05.p01~qt~vecgeom~opengl~x11~motif~data~clhep+threads
-    - $root
-    - geant3@2-7_fairsoft
-    - vgm@4-5
-    - geant4-vmc@4-0-p1
-    - fairroot@18.4.0+sim+examples
+  - $concretizer_hints
+  - fairsoft-config@jun19
```
ChristianTackeGSI, line 14: As the `jun19_fairroot_18_4` env is mainly for CI (maybe should be moved to `test/`?), why do you need to add fairsoft-config? Or does that fairoot-Version (18.4) really need it?

env/dev/r3broot/spack.yaml
dennisklein: > why do you need to add fairsoft-config?

FairRoot depends on it, I am just pinning the version to `jun19`.

ChristianTackeGSI: I am not sure, I like the "fairroot depends_on fairsoft-config".
* On the one hand, I can understand it.
* On the other hand: fairoot itself doesn't need it anymore, correct? So only the envs for expirements need it.

dennisklein: (Technically it does not need to depend on it, but if it is not, I cannot use `spack view --dependencies true -e fairroot dir... fairroot`.)

dennisklein: > As the r3broot env is mainly for CI (maybe should be moved to test/?

Sure, I don't mind where it is.

dennisklein: > (Technically it does not need to depend on it, but if it is not, I cannot use `spack view --dependencies true -e fairroot dir... fairroot`.)

Or, is there another way to attach a view to an env and exclude `fairroot` without modifying the env?

ChristianTackeGSI: > > (Technically it does not need to depend on it, but if it is not, I cannot use `spack view --dependencies true -e fairroot dir... fairroot`.)
> 
> Or, is there another way to attach a view to an env and exclude `fairroot` without modifying the env?

I don't know, if this is an option, but what about adding the (relevant) packages as `depends_on()` to fairsoft-config? Like "If you install fairsoft-config, you expect the legacy fairsoft packages to be installed somewhere"?

dennisklein: > but what about adding the (relevant) packages as depends_on() to fairsoft-config?

Because it will be exactly === all the run-time dependencies of the fairroot pkg.

thisfairsoft.sh
dennisklein: Done, see #298

dennisklein:
I split the gcc10 fixes into a separate PR #298.

env/dev/r3broot/spack.yaml
dennisklein: we concluded to keep it as dependency of fairroot offline.

env/dev/jun19_fairroot_18_4/spack.yaml
dennisklein: We can move the envs to test in a separate PR.

repos/fairsoft/packages/fairroot/package.py
```diff
@@ -6,96 +6,76 @@
...
     patch('fairlogger_incdir.patch', level=0, when='@18.2.1')
+    patch('find_pythia8_cmake.patch', when='@:18 +sim')
 
     def setup_build_environment(self, env):
         super(Fairroot, self).setup_build_environment(env)
-        stdversion=('-std=c++%s' % self.spec.variants['cxxstd'].value)
-        env.append_flags('CXXFLAGS', '-std=c++%s' % self.spec.variants['cxxstd'].value)
+        stdversion = ('-std=c++%s' % self.spec.variants['cxxstd'].value)
+        env.append_flags('CXXFLAGS',
+                         '-std=c++%s' % self.spec.variants['cxxstd'].value)
```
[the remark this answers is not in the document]
dennisklein: removed it

repos/fairsoft/packages/fairroot/package.py
```diff
@@ -6,96 +6,76 @@
...
-    depends_on('pythia6', when='+sim')
-    depends_on('pythia8', when='+sim')
-
-    # mesa and libxml2 are dependencies of root which have to be build extra due to the
-    # extra build options
-    depends_on('libxml2~python')
-
-    # Dependencies for dev version
+    depends_on('boost@1.68.0: +container')
     depends_on('boost@1.68.0: cxxstd=11 +container')
```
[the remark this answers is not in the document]
dennisklein: removed it

test/buildenv.sh
```diff
@@ -28,6 +28,33 @@ then
 	spack install
 	ret=$?
 fi
+if [ "$ret" = 0 -a -n $2 ]
```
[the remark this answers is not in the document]
dennisklein: done

CMakeLists.txt
```diff
@@ -90,25 +90,30 @@ set_property(TEST test.start APPEND PROPERTY
...
+fs_test(NAME test.odc                 SPEC odc)
+fs_test(NAME jun19.sim                ENV env/jun19/sim)
+fs_test(NAME jun19.sim_mt             ENV env/jun19/sim_mt)
+fs_test(NAME dev.sim                  ENV env/dev/sim)
+fs_test(NAME dev.sim_mt               ENV env/dev/sim_mt)
+fs_test(NAME dev.sim_mt_headless      ENV env/dev/sim_mt_headless)
+fs_test(NAME test.jun19_fairroot_18_4 ENV test/env/jun19_fairroot_18_4)
+fs_test(NAME test.fairroot_develop    ENV test/env/fairroot_develop)
+
+fs_test(NAME dev.r3broot              ENV env/dev/r3broot)
```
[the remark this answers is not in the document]
dennisklein: done

repos/fairsoft/packages/fairroot/package.py
```diff
@@ -6,96 +6,76 @@
...
         options.append('-DBOOST_INCLUDEDIR={0}/include'.format(
-        self.spec['boost'].prefix))
+            self.spec['boost'].prefix))
         options.append('-DBOOST_LIBRARYDIR={0}/lib'.format(
-        self.spec['boost'].prefix))
+            self.spec['boost'].prefix))
         options.append('-DFlatbuffers_DIR={0}'.format(
-        self.spec['flatbuffers'].prefix))
+            self.spec['flatbuffers'].prefix))
         options.append('-DDISABLE_GO=ON')
```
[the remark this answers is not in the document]
dennisklein: removed it

CMakeLists.txt
```diff
@@ -90,25 +90,29 @@ set_property(TEST test.start APPEND PROPERTY
...
+fs_test(NAME test.fairmq              ENV test/env/fairmq)
+fs_test(NAME test.odc                 SPEC odc)
+fs_test(NAME jun19.sim                ENV env/jun19/sim)
+fs_test(NAME jun19.sim_mt             ENV env/jun19/sim_mt)
+fs_test(NAME dev.sim                  ENV env/dev/sim)
+fs_test(NAME dev.sim_mt               ENV env/dev/sim_mt)
+fs_test(NAME dev.sim_mt_headless      ENV env/dev/sim_mt_headless)
+fs_test(NAME test.jun19_fairroot_18_4 ENV test/env/jun19_fairroot_18_4)
+fs_test(NAME test.fairroot_develop    ENV test/env/fairroot_develop)
+fs_test(NAME test.r3broot             ENV env/dev/r3broot)
```
ChristianTackeGSI, line 107:
```suggestion
fs_test(NAME test.r3broot             ENV test/env/r3broot)
```

ChristianTackeGSI (approved):

CMakeLists.txt
dennisklein: done

dennisklein:
If tests are still broken I will fix em in a separate PR.

PR #284, opened 2020-08-06
VAE Split: keep things for VAE, move to spack 0.14.2

ChristianTackeGSI:
This is a bit bigger:

* Give VAE its own backports spack repo
* Give VAE its own spack (copied from current)
* Give VAE its own `vae/thisvae.sh` (adapted from `thisfairsoft.sh`)
* Update main spack to 0.14.2
  Yes, 0.15.x is already out there and we want to go there. This is a short intermediate / refuelling stop: In the past, each spack upgrade meant some breakage. So we do this in digestible steps.
* Move retired backports to `vae-backports`. Some of our backports are already usable in 0.14.x. So move those recipes to `vae-backports` for VAE to use them.
* Retire `pcre+jit` concretizer hint. It's not needed with 0.14.x either. If VAE needs it, we should create a `vae/config/packages.yaml` that hints the `+jit` for pcre there.


ChristianTackeGSI:
I have no direct idea, why openblas fails NOW.
dev_new_spack used to work for 0.14.x.
I am rerunning the old dev_new_spack for now. Let's see, how this works out.
If that fails also, it's something different on our CI (maybe macos update?)
If it runs fine, then something intermediate broke it. Then I will slowly rebase dev_new_spack to find out (and possibly modify the testsuite to only build openblas)

PR #286, opened 2020-08-06
CI: Increase passed test output size, Misc

ChristianTackeGSI:
On some cmake versions, the output of passed tests is truncated to 1024 Bytes. That's really not much. At least we would like to see the concretizer output! So increased this to 200k.

Let each test output the spack version that is being used. We should know it for each test. But better be verbose, as we're starting to mix versions.

And yet another round of reformatting output.


ChristianTackeGSI:
Faling test seems to be the fairmq-issue.

PR #304, opened 2020-08-25
g4*: Update version lines

ChristianTackeGSI:
* Add `sha256=` prefix everywhere
* Reorder: Most recent version at top
* Update copyright year


dennisklein (approved):
I did not verify all those hash sums ;)

PR #306, opened 2020-08-27
Make thisfairsoft more "fail safe"

ChristianTackeGSI:
thisfairsoft used to do some things that are good for initial setup, or for fixing some small problems, like
* Call git submodule update --init
* Call spack clean

Those commands are neither "read only", nor safe under all circumstances. Especially they will break an ongoing `spack install`.

So switch model a bit to "fail safe":
* `thisfairsoft --setup` gives the old behaviour.
  As noted, this is good for setup and fixing problems
* `thisfairsoft` does not modify anything, but gives warnings, if things are not correctly configured and a `--setup` run is needed.

If a user misses the `--setup` step entirely, they will not have the fairsoft spack repo, and so many `spack install`s will just fail, without breaking anything.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,10 +113,10 @@ fs_test(NAME test.r3broot             ENV test/env/r3broot)
   # fs_test(NAME test.openmpi_3 ENV test/env/openmpi_3 MIGHT_FAIL)
 # endif()
 
-fs_test(NAME workflow.classic_developer_jun19 ENV env/jun19/sim DEPENDS jun19.sim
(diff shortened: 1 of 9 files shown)


dennisklein (approved):

dennisklein:
https://cdash.gsi.de/testDetails.php?test=7094992&build=245315 -> `Error: Timed out waiting for a read lock.`

@ChristianTackeGSI Didn't you have a fix lying around for that already? Can you push it here?

dennisklein:
just rebased, tests were green on last run.

PR #310, opened 2020-08-31
Fix workflow tests

dennisklein:


test/workflow/classic_developer.sh
```diff
@@ -3,17 +3,33 @@
...
 unset FAIRSOFT_ROOT
 export SIMPATH="$(realpath ./fairsoft)"
 cmake="$SIMPATH/bin/cmake"
-spack view --dependencies true -e fairroot symlink -i $SIMPATH fairroot cmake
+ctest="$SIMPATH/bin/ctest"
+
+unset extra_excludes
+if [ "$(uname)" = Darwin ]
+then
+  extra_excludes="-e libpng -e libjpeg-turbo -e libiconv -e sqlite"
```
ChristianTackeGSI, line 19: I start to wonder, what would happen, if we do that on Linux as well…
A real problem could arise, when an experiment (or FairRoot in this case) wants to link to one of the excluded packages.

dennisklein: > I start to wonder, what would happen, if we do that on Linux as well…

I think it is already dirty to do this on macOS^^

IMHO, the real solution is to not model these dependencies at all in spack, if we filter them out anyways afterwards (and partially only, because we still keep them referenced via RPATH ... how are you going to explain this to anyone?)

dennisklein: It would be a new type of dependency: `link` minus `install-in-certain-cases-where-we-use-it-the-wrong-way` ...

ChristianTackeGSI (approved):
Looks good

PR #312, opened 2020-09-01
CI: Enable workflows, FS_TEST_INSTALLTREE

ChristianTackeGSI:
* Enable all workflows again. Yes, they might break on mac. Let's see.
* To be able to customize install tree, introduce FS_TEST_INSTALLTREE cmake/env var.


dennisklein (approved):

dennisklein:
Do not have a good idea on the test timeout, I'll merge and we will see on the full `dev` build again.

PR #315, opened 2020-09-02
Enable install tree caching on our macs

ChristianTackeGSI:
First, implement the uninstall part

Second, write ths git sha1 of the currently building version into a stamp file in the install-tree, so that we have something to compare against at the next run.

* If the run is starting (and thus unknown), then the sha1 is appended, as, the tree might still reflect something from the previous run(s).
* If the run is successful, then replace the complete contents with the current sha1


dennisklein (approved):

dennisklein:
The Ubuntu build failures are unrelated (no space left).

PR #320, opened 2020-09-14
Move cluster-related part to vaesoft subdirectory.

kresan:
This change will allow to run `git filter-branch` in one go on the vaesoft subdirectory.


dennisklein (approved):

kresan:
FairMQ build fails due to error in PMIxPlugin. @dennisklein Could you please have a look whether it is related to my changes? I do not expect that, but maybe I missed something.

dennisklein:
> FairMQ build fails due to error in PMIxPlugin. @dennisklein Could you please have a look whether it is related to my changes? I do not expect that, but maybe I missed something.

It is unrelated.

PR #322, opened 2020-09-22
New FairMQ versions and XCode 12 fix

dennisklein:
* Backport fix for FairRootGroup/FairMQ#298
* Package new versions
  * https://github.com/FairRootGroup/FairMQ/releases/tag/v1.4.23
    * Needed a patch for FairRootGroup/FairMQ#307
  * https://github.com/FairRootGroup/FairMQ/releases/tag/v1.4.24


repos/fairsoft/packages/fairmq/package.py
```diff
@@ -57,6 +59,8 @@ class Fairmq(CMakePackage):
     patch('fix_find_dds.patch', when='@1.4.0:1.4.4')
     patch('use_bundled_gtest_149.patch', when='@1.4.9:1.4.16')
     patch('missing_stdexcept_header.patch', when='@:1.4.19')
+    patch('fix_cpp17moveinsertable_assertion_xcode12.patch', when='@:1.4.23')
+    patch('update_command_format_in_pmix_plugin.patch', when='@1.4.23')
```
ChristianTackeGSI, line 63: I *personally* think, that we don't need to package each and every version. Especially, if 1.4.24 could be considered a "hot fix" for 1.4.23?

ChristianTackeGSI (approved):
Looks good. Maybe the CI should be triggered again at some point?

repos/fairsoft/packages/fairmq/package.py
dennisklein: > Especially, if 1.4.24 could be considered a "hot fix" for 1.4.23

it is slightly more.

> I personally think, that we don't need to package each and every version.

I agree, but for FairMQ I will do the work as an exception :)

dennisklein:
> Maybe the CI should be triggered again at some point?

It is running still, just the yellow status light were overridden by the previous build which i stopped manually after the second run already started.

PR #333, opened 2020-10-05
Port legacy install scripts to CMake

dennisklein:
Benefits:
* much shorter and more readable
* models package dependencies
* checksum support for plaintext downloads
* ~~new dialog based `configure.sh`~~
* bootstrap CMake (if too old) with binary download (for testing I require 3.18, will relax this later)
* full out-of-source build support

Standard CMake workflow:
```
git clone -b legacy https://github.com/dennisklein/FairSoft
cmake -S FairSoft -B FairSoft_build -C FairSoft/FairSoftConfig.cmake
cmake --build FairSoft_build -j 8
```

With the above you will choose an out-of-source build and install directory. If the build aborts for some reason, continue/re-run by typing `make` in the chosen build directory.

Motivation: The goal for the standard workflow is that I want to be able to write standard CTests to be able to test also legacy installations with our existing infrastructure.

---

~~[Demo recording](https://asciinema.org/a/9efr2ttudZDcabzy9CtLrOww6)~~

FairSoft_1

FairSoft_2

FairSoft_3

Hackable by anyone:
[`cmake/legacy.cmake`](https://github.com/FairRootGroup/FairSoft/blob/30fd77e4bf42848e4d7b10d12de11c8ce2dc446f/cmake/legacy.cmake)
[`configure.sh`](https://github.com/FairRootGroup/FairSoft/blob/30fd77e4bf42848e4d7b10d12de11c8ce2dc446f/configure.sh) (will probably move to `legacy/configure.sh`)

Full integration with alfa-ci (e.g. [Jenkins](https://alfa-ci.gsi.de/blue/organizations/jenkins/FairRootGroup%2FFairSoft/detail/PR-333/70/pipeline) and [CDash](https://cdash.gsi.de/index.php?project=FairSoft))

---

TODO
* [x] Solve macOS Pythia8 problem
* [x] Update dialog menu
* [x] Docs
* [x] Add tests for all flavors: Fedora, CentOS, Ubuntu, Debian, OpenSuse, macOS


FairSoft_test.cmake
```diff
@@ -65,24 +68,25 @@ if (USE_TEMPDIR)
     message(STATUS "${timestamp} ... done")
 else()
     # No temporary directory created
-    set(FS_TEST_WORKDIR "")
+    set(FS_TEST_WORKDIR ".")
```
ChristianTackeGSI, line 71: I don't like that one, because `"."` is not correct in that case.

If `FS_TEST_WORKDIR` is empty/unset,  `build/test_workdir` is used.
```suggestion
    set(FS_TEST_WORKDIR "${CTEST_BINARY_DIRECTORY}/test_workdir")
```

FairSoft_test.cmake
```diff
@@ -65,24 +68,25 @@ if (USE_TEMPDIR)
...
-    if ("${FS_TEST_WORKDIR}" STREQUAL "")
-        get_filename_component(FS_TEST_INSTALLTREE
-                               "${CTEST_BINARY_DIRECTORY}/install-tree"
-                               ABSOLUTE)
-    else()
-        set(FS_TEST_INSTALLTREE "${FS_TEST_WORKDIR}/install-tree")
-    endif()
+if (NOT BUILD_METHOD STREQUAL legacy)
+  if ("${FS_TEST_INSTALLTREE}" STREQUAL "")
+      if ("${FS_TEST_WORKDIR}" STREQUAL "")
```
ChristianTackeGSI, line 77:
```suggestion
      if ("${FS_TEST_WORKDIR}" STREQUAL "${CTEST_BINARY_DIRECTORY}/test_workdir")
```

FairSoft_test.cmake
```diff
@@ -134,20 +138,24 @@ file(REMOVE "${CTEST_BINARY_DIRECTORY}/CMakeCache.txt")
 # Remove workdir (keep in sync with CMakeLists.txt)
 file(REMOVE_RECURSE "${CTEST_BINARY_DIRECTORY}/test_workdir")
```
ChristianTackeGSI, line 139:
```suggestion
file(REMOVE_RECURSE "${FS_TEST_WORKDIR}")
file(MAKE_DIRECTORY "${FS_TEST_WORKDIR}")
```

Jenkinsfile
```diff
@@ -27,13 +39,17 @@ def jobMatrix(String node_type, String ctestcmd, List specs, Closure callback) {
         try {
           checkout scm
 
+          if (isLegacyChange()) {
+            ctestcmd = ctestcmd + ' -DBUILD_METHOD=legacy'
+          }
+
```
ChristianTackeGSI, line 45: Could you look into moving that outside the jobMatrix function?
Maybe after the `def ctestcmd = "ctest -VV -S FairSoft_test.cmake"`?

cmake/spack.cmake
```diff
@@ -0,0 +1,125 @@
...
+find_package(Git)
+if(GIT_FOUND)
+  execute_process(COMMAND ${GIT_EXECUTABLE} submodule update --init
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
+endif()
+
+# Test #######################################################################
+include(CTest)
+
+if (DEFINED FS_TEST_WORKDIR AND NOT "${FS_TEST_WORKDIR}" STREQUAL ".")
```
ChristianTackeGSI, line 18:
```suggestion
if (DEFINED FS_TEST_WORKDIR AND NOT "${FS_TEST_WORKDIR}" STREQUAL "")
```

ChristianTackeGSI (commented):
Especially the FS_TEST_WORKDIR stuff is UNTESTED.
It's just my idea on how this might work better.

Jenkinsfile
dennisklein: Not sure how, it needs to run after the scm checkout and this only happens in node context. I could write it inside the lambda on the callsite.

ChristianTackeGSI: Ohh, if it needs the scm checkout, then all is fine. Maybe add a quick comment to that effect here?

FairSoft_test.cmake
```diff
@@ -123,31 +133,41 @@ if (NOT "$ENV{BRANCH_NAME}" STREQUAL "")
...
-file(REMOVE_RECURSE "${CTEST_BINARY_DIRECTORY}/test_workdir")
-
-
-ctest_start(Continuous TRACK ${cdash_group})
-ctest_configure(OPTIONS "-DFS_TEST_WORKDIR=${FS_TEST_WORKDIR};-DFS_TEST_INSTALLTREE=${FS_TEST_INSTALLTREE}")
-# ctest_submit(PARTS Start Configure)
-
-# ctest_build()
+if (BUILD_METHOD STREQUAL legacy)
+  ctest_empty_binary_directory(${CTEST_BINARY_DIRECTORY})
```
ChristianTackeGSI, line 137: on macOS, the (to be cached) install-tree is in `build/install-tree`. ;)
So this might kill the install-tree.

I think, I should write up a "test architecture" docs section. ;-)

ChristianTackeGSI: *grrr* I missed the `if (BUILD_METHOD STREQUAL legacy)` above it.

FairSoft_test.cmake
dennisklein: I reverted to the old way (empty)

cmake/spack.cmake
dennisklein: done

FairSoft_test.cmake
dennisklein: I reverted to the old way (empty)

FairSoft_test.cmake
dennisklein: ya, i suspected something like this. any arguments against moving `install-tree` besides the build dir?

kresan:
Have tried it. Looks very user-friendly.
Does it make sense to have automatic `cd %BUILD_DIR% && make` after running the configure.sh?

dennisklein:
I had it like this before, certainly doable. One reason I tried it as two steps is to show the user that it is a proper cmake workflow now and enable them to also realize the configure.sh is just optional. For the first invocation having it called automatically is certainly much nicer. But what if the build does not run through because something was missing, then user corrects it and runs again how? By calling configure.sh again or by typing `make` in the build dir? This i have not thought to the end yet.

dennisklein:
The CentOS 7 and openSUSE 15.2 builds succeeded, just the cdash upload failed.

PR #337, opened 2020-10-09
Update docs for the upcoming mixed legacy/spack release

ChristianTackeGSI:
* Made the main README into a "dispatch" function: Choose your method and continue in another place
* Split up the sections of the spack docs into individual files
* Update the stuff with the status information

Instead of reading the diff, I would recommend to look at the result:
https://github.com/ChristianTackeGSI/FairSoft/blob/pr/docs/README.md

We can either merge this quickly and improve from there or improve the docs part in my pr/docs branch. I am fine with both.


legacy/README.md
```diff
@@ -1,4 +1,17 @@
...
+## Preface
+
+This will be the last release featuring the bash/cmake
+based setup system. Upcoming releases are planned to use
+spack instead (for more information see
+[here](../docs/README.md)).
+
+The remainder of this README is basically the main README
+from the last releases. All file references should be read
+as relative to the `/legacy/` folder.
```
dennisklein, line 12:
```suggestion
Our homegrown bash/cmake based setup system
has been named "Legacy". It has been moved to the
sub-directory `legacy/` to distinguish it clearly
from the future Spack-based setup system
(for more information see [here](../docs/README.md)).
The latter will eventually replace the "Legacy" setup system
in a future release.
```

* I prefer to advertize the upcoming spack-only release "less final". Sure, right now it looks like it is going to be the next release in half a year. But what if we have to release another time in between, which we do not know yet today.
* Let's delete the note on references and rather update such references everywhere please.

dennisklein (changes requested):
> We can either merge this quickly and improve from there or improve the docs part in my pr/docs branch. I am fine with both.

Let's merge rather fast and go from there. But please consider my inline comment now.

README.md
```diff
@@ -2,298 +2,14 @@
...
-`docs/` contains additional documentation pages, the relevant ones for the user should all be linked from this top-level page.
-
-`legacy/` contains the former shell script based FairSoft before we switched to Spack, it is no longer updated, but remains for reference.
-
-`test/` contains mainly integration test related files used by the FairSoft Continuous Integration checks. The normal user can ignore them.
-
-CMake/CTest-related files are mainly for internal use and running the FairSoft Continuous Integration checks. The normal user can ignore them.
+|  **[Legacy](legacy/README.md)** | **[Spack](docs/README.md) (EXPERIMENTAL)** |
+| -- | -- |
+| This is the homegrown bash/cmake based setup system. | This is an ongoing standardization and modernization effort based on Spack (which itself is still under heavy development). Most things are already working. For early adopters. |
```
dennisklein, line 11:
```suggestion
| This is the classic bash/cmake based setup system. | This is an ongoing standardization and modernization effort based on Spack (which itself is still under heavy development). Most things are already working. For early adopters. |
```

ChristianTackeGSI:
Added requested change.

Also increased quietPeriod a lot.

legacy/README.md
```diff
@@ -31,54 +43,55 @@ This Release was tested on following systems:
...
 
-  FairSoft> ./configure.sh
-
-In the first menu one has to define the compiler which should be used for the
-installation. The normal choices would be _gcc_ for Linux and _Clang_ for Mac OSX.
-The compiler must be installed on the system. If one choose a compiler which is not
-installed, the installation procedure will stop when checking for the prerequisites
-defined above.
+```
+$ ./configure.sh
```
dennisklein, line 51:
```suggestion
legacy/$ ./configure.sh
```

legacy/README.md
```diff
@@ -31,54 +43,55 @@ This Release was tested on following systems:
 ## Guided Installation
 
 To start the installation procedure one has to run the configure.sh script which is
-found in the main directory of FairSoft from within this directory. If the script is
+found in the current directory of FairSoft from within this directory. If the script is
```
dennisklein, line 46:
```suggestion
found in the `legacy/` directory of FairSoft. If the script is
```

dennisklein (changes requested):
Let's make these references more explicit please. Have not looked through the whole document. Please find all of them.

dennisklein (approved):

PR #340, opened 2020-10-21
dds: Fix patch for @develop, add version 3.5.{2,3}

ChristianTackeGSI:
* `@develop` (master) changed a bit, so the patch needed an update
* In #333, Dennis used dds at version 3.5.2. So add it to spack as well.
* While at it, add version 3.5.3.


ChristianTackeGSI:
The CI failures are related to workflow.

Open question: Should I add 3.5.3?
(I want to have the same versions, that any legacy thing can build!)

ChristianTackeGSI:
Still only the workflow tests fail.

Seems unrelated.

I will look at the wokflow tests next.

dennisklein (approved):

Issue #344, opened 2020-11-02
Install also ODC for release nov20


fuhlig1:
Since some projects already use ODC or want to use it in near future it would be good if ODC is installed with all the other FairSoft packages. The installation would reduce the work for users.

dennisklein:
@AndreyLebedev Which version of ODC should we best include in the `nov20` release of FairSoft? Do you have any comments, remarks regarding the compile config and/or needed dependencies?

AndreyLebedev:
@dennisklein The latest version is 0.10 is the one which supports multiple partitions (DDS sessions). I would recommend using it. But it requires the latest DDS (3.5.3) and the latest FairMQ (1.4.26) versions. If this is not an issue then please use the latest.

ChristianTackeGSI:
For FairSoft spack edition, see #345. Some questions still open, see there.

dennisklein:
Implemented.

PR #345, opened 2020-11-02
Add odc@0.10, fairmq@1.4.{25,26}, fairlogger@1.9.0

ChristianTackeGSI:
#344 asks for odc@0.10, so let's add it.
It also names new deps for odc.

We should discuss, how to handle 0.8:
* I have used the upgraded deps for 0.8 as well currently. This is probably unneeded.
* We could correct this and give correct deps
* As nobody should use odc@0.8, we could also just drop it! (I'd vote for this. But I can easily convinced to do otherwise.)

And we should clear up, if we want to add odc@0.10 (or odc at all) to the `dev` environments?


AndreyLebedev:
> * As nobody should use odc@0.8, we could also just drop it! (I'd vote for this. But I can easily convinced to do otherwise.)

I think we can easily drop 0.8, since nobody except ALICE uses ODC for the moment. ALICE uses `alibuild` anyway.

ChristianTackeGSI:
Okay, dropped odc@0.8 (CI just started)

Which leaves the question:
Do we want odc@0.10 in the `dev` spack environments?
@fuhlig1, @kresan do you have any opinion?

AndreyLebedev:
> Do we want odc@0.10 in the `dev` spack environments?

In general, there is nothing against it. Spack installation of`ODC` will mainly be used by `CBM`. If `CBM` needs it then the answer is `yes`.

fuhlig1:
I can't tell you yet if we will need the spack installation in near future but I would anyway propose to add odc to the spack environments.

dennisklein:
>Do we want odc@0.10 in the dev spack environments?

Since it is a top-level packages no other package depends on currently, I am fine adding it to the environments. But please add it without any explicit version. It is a non-physics-sensible package which is still under development, so we anyway will want the newest releases in the forseeable future (similar to FairMQ and DDS).

ChristianTackeGSI:
Failing tests are workflow related. Still working on those in another PR.

dennisklein (approved):

PR #346, opened 2020-11-11
yaml-cpp: shared library building; msgpack-c: checksums

ChristianTackeGSI:
* yaml-cpp: Backport from upstream (shared library)

  Fix shared library building of yaml-cpp

  See: https://github.com/FairRootGroup/FairSoft/commit/7778fcb23bd525be18cfd2
  See: https://github.com/spack/spack/pull/19866

* msgpack-c: Switch checksums to sha256

### DIFF
--- a/repos/fairsoft-backports/packages/msgpack-c/package.py
+++ b/repos/fairsoft-backports/packages/msgpack-c/package.py
@@ -11,10 +11,10 @@ class MsgpackC(CMakePackage):
     homepage = "http://www.msgpack.org"
     url      = "https://github.com/msgpack/msgpack-c/archive/cpp-3.0.1.tar.gz"
 
-    version('3.1.1', '977c85b4f76c7b8d28e348552713d8ef')
-    version('3.0.1', 'a79f05f0dc5637c161805d6c0e9bfbe7')
-    version('2.1.5', '6536e2072a1006e2004e2963081692a2')
-    version('1.4.1', 'e2fd3a7419b9bc49e5017fdbefab87e0')
+    version('3.1.1', sha256='bda49f996a73d2c6080ff0523e7b535917cd28c8a79c3a5da54fc29332d61d1e')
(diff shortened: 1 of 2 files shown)


ChristianTackeGSI:
The current CI failures are all "workflow"-related, which we handle in another PR.

dennisklein (approved):
For earlier versions, this will create a CMake warning about an unused variable for `YAML_BUILD_SHARED_LIBS`, but I guess this should be fine.

ChristianTackeGSI:
> For earlier versions, this will create a CMake warning about an unused variable for `YAML_BUILD_SHARED_LIBS`, but I guess this should be fine.

If this has a potential of breaking something at some point, I'd be inclined to fix it. But I don't see that potential anywhere. So I'd prefer to keep the recipe more simple.

Or do you see any potential there?

dennisklein:
> Or do you see any potential there?

Just a -Werror equivalent for CMake if there is such a thing.

ChristianTackeGSI:
Hmm, that exists!

Looks like it is `-Werror=dev`.
See: https://gitlab.kitware.com/cmake/cmake/-/commit/28f2d750edaf6ee1af660d3a0ae6792c65c47997

We maybe should add this to some of our CI-jobs?

I am not sure, whether I would like to see this in the spack recipes for _our_ packages. On one side, it would catch wrong recipes. But on the other side it would likely make things more unstable for our users.

dennisklein:
> We maybe should add this to some of our CI-jobs?

`-Werror=deprecated` is maybe interesting, `-Werror=dev` I don't know, I personally don't need warnings to be errors, I would rather make it a good review practice to also check warnings and try to fix/avoid them if possible.

In any case, 
* I would not enable such things for the user, at maximum in our CI and experimental builds!
* we are not testing against a complete range of supported CMake versions, some versions might not warn at all or in case of deprecation warnings a fix could break compatibility with older supported CMake versions?!

PR #352, opened 2020-11-23
Sync spack with latest legacy nov20

dennisklein:


dennisklein:
spack CI builds run again, but they need much more work. Ignoring it for the purpose of this PR.

PR #355, opened 2020-12-11
[spack] dds: Update patches, add 3.5.4

ChristianTackeGSI:
* dds@develop (master) had some changes, that required our patches to be updated
* macos 11 has strange libraries, that don't exist.
* Add Version 3.5.4 (allows newer boost version)


ChristianTackeGSI:
macos failing:
* jun19 is failing due to old geant3, fix in another PR
* workflow-dev is still in "expected" state.

dennisklein:
Regarding 1f21e66, did you try that with CMake >= 3.18.0? Upstream appears to have addressed it: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4998

ChristianTackeGSI:
> Regarding [1f21e66](https://github.com/FairRootGroup/FairSoft/commit/1f21e6636959d068a8794379fddf9d2c8268ce64), did you try that with CMake >= 3.18.0? Upstream appears to have addressed it: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4998

Thanks for the hint!
* Backported cmake
* Dropped 1f21e66
* Updated `depends_on` (maybe we should update the main one as in "We need 3.18 for reliable GET_RUNTIME_DEPENDENCIES support?)

Let's see.

ChristianTackeGSI:
Only pixelSplitDDS seems to fail.

Do we want to upgrade legacy to 3.5.4 as well?

Do we want to update odc to 0.12 on spack and legacy?

ChristianTackeGSI:
Having just learned, that dds 3.5.4 requires C++17, I wonder, whether I should add
```python
conflicts('cxxstd=11', when='@3.5.4:')
conflicts('cxxstd=14', when='@3.5.4:')
```
to the recipe?

ChristianTackeGSI:
On Linux only the workflow tests fail. (pixelSplitDDS again).

On macOS something about opengl is breaking. Not related either.

Ready to be reviewed (again).

PR #358, opened 2020-12-16
Update setup-debian.sh

laurentaudouin:
rsync required during compilation
added a line for an updated version of cmake (3.16 is mandatory despite being a very recent version of cmake)

### DIFF
--- a/legacy/dependencies.md
+++ b/legacy/dependencies.md
@@ -5,7 +5,7 @@ you may need to run the script with `sudo`.
 
 * CentOS 7:
   [`legacy/setup-centos-7.sh`](setup-centos-7.sh)
-* Debian 10: Please highly consider enabling [packports](https://backports.debian.org/):
+* Debian 10: Please highly consider enabling [backports](https://backports.debian.org/):
   [`legacy/setup-debian.sh`](setup-debian.sh)
 * Fedora:
   [`legacy/setup-fedora.sh`](setup-fedora.sh)
--- a/legacy/dependencies.md
+++ b/legacy/dependencies.md
@@ -3,12 +3,18 @@
 Pick the appropriate setup script for your system and run it. On Linux systems
 you may need to run the script with `sudo`.
 
-* [`legacy/setup-centos-7.sh`](setup-centos-7.sh)
-* [`legacy/setup-debian.sh`](setup-debian.sh)
-* [`legacy/setup-fedora.sh`](setup-fedora.sh)
-* [`legacy/setup-macos.sh`](setup-macos.sh)
-* [`legacy/setup-opensuse.sh`](setup-opensuse.sh)
-* [`legacy/setup-ubuntu.sh`](setup-ubuntu.sh)
+* CentOS 7:
+  [`legacy/setup-centos-7.sh`](setup-centos-7.sh)
+* Debian 10: Please highly consider enabling [packports](https://backports.debian.org/):
+  [`legacy/setup-debian.sh`](setup-debian.sh)
+* Fedora:
+  [`legacy/setup-fedora.sh`](setup-fedora.sh)
+* macOS:
+  [`legacy/setup-macos.sh`](setup-macos.sh)
+* OpenSuse:
+  [`legacy/setup-opensuse.sh`](setup-opensuse.sh)
+* Ubuntu:
+  [`legacy/setup-ubuntu.sh`](setup-ubuntu.sh)
 
 **Please read the scripts carefully and make sure you agree with the changes to your system!!**
 
--- a/legacy/setup-debian.sh
+++ b/legacy/setup-debian.sh
@@ -4,11 +4,12 @@ export DEBIAN_FRONTEND=noninteractive
 apt-get update
 apt-get -y upgrade
 apt-get -y install autoconf automake binutils \
-bison build-essential bzip2 ca-certificates cmake cmake-data coreutils \
+bison build-essential bzip2 ca-certificates cmake-data coreutils \
 curl debianutils file findutils flex g++ gcc gfortran git gzip \
 hostname libbz2-dev libcurl4-openssl-dev libicu-dev libgl1-mesa-dev libglu1-mesa-dev \
 libncurses-dev libssl-dev libtool libx11-dev libxerces-c-dev libxext-dev libxft-dev \
 libxml2-dev libxmu-dev libxpm-dev libyaml-cpp-dev lsb-release make patch python-dev \
-python3-dev sed subversion tar unzip wget xutils-dev xz-utils
+python3-dev rsync sed subversion tar unzip wget xutils-dev xz-utils
+apt-get -t buster-backports install cmake
 apt-get -y clean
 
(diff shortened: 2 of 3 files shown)


legacy/setup-debian.sh
```diff
@@ -4,11 +4,12 @@ export DEBIAN_FRONTEND=noninteractive
...
 apt-get -y install autoconf automake binutils \
-bison build-essential bzip2 ca-certificates cmake cmake-data coreutils \
+bison build-essential bzip2 ca-certificates cmake-data coreutils \
 curl debianutils file findutils flex g++ gcc gfortran git gzip \
 hostname libbz2-dev libcurl4-openssl-dev libicu-dev libgl1-mesa-dev libglu1-mesa-dev \
 libncurses-dev libssl-dev libtool libx11-dev libxerces-c-dev libxext-dev libxft-dev \
 libxml2-dev libxmu-dev libxpm-dev libyaml-cpp-dev lsb-release make patch python-dev \
-python3-dev sed subversion tar unzip wget xutils-dev xz-utils
+python3-dev rsync sed subversion tar unzip wget xutils-dev xz-utils
+apt-get -t buster-backports install cmake
```
dennisklein, line 13: Doesn't this also require some config to enable the backports archives?

In any case, if we hard-code `buster`, let's duplicate the file as `setup-debian-10.sh`.

dennisklein (changes requested):
Thx for the contrib! Please see my inline question and comment.

dennisklein:
> ... (3.16 is mandatory despite being a very recent version of cmake)

Just for context: The main reason we require this *recent* CMake version, is that some of the packages still rely on the old `FindBoost.cmake` module which is shipped with CMake (rather than using the newer CMake package shipped with Boost) to discover Boost in their CMake-based build systems. While older CMake versions might work, only CMake 3.16's `FindBoost.cmake` properly supports discovery of Boost 1.72.

legacy/setup-debian.sh
laurentaudouin: This commands for backports works out of the box (to my own surprise :) ).
And yes, it's wise to mention it's for debian 10 only
Note that the rsync requirement is generic, though.

ChristianTackeGSI: Nope, backports need to be enabled for this to work:
```console
root@lxci002:~# lsb_release -d
Description:    Debian GNU/Linux 10 (buster)
root@lxci002:~# apt-get -t buster-backports install cmake
Reading package lists... Done
E: The value 'buster-backports' is invalid for APT::Default-Release as such a release is not available in the sources
```

So if you do this (which is okay with me), please consider:
* Renaming setup-debian to setup-debian-10. We support Debian in version 10 only, currently. (and check the references to it, and update them as needed).
* Checking the docs, and add a note about backports should be enabled. Possibly linking to backports.debian.org.

legacy/setup-debian.sh
```diff
@@ -4,11 +4,12 @@ export DEBIAN_FRONTEND=noninteractive
 apt-get update
 apt-get -y upgrade
 apt-get -y install autoconf automake binutils \
-bison build-essential bzip2 ca-certificates cmake cmake-data coreutils \
+bison build-essential bzip2 ca-certificates cmake-data coreutils \
```
ChristianTackeGSI, line 7:
```suggestion
bison build-essential bzip2 ca-certificates coreutils \
```
Just remove cmake completely. cmake-data is pulled in by cmake anyway.

ChristianTackeGSI:
Hi @laurentaudouin,

I would like to continue on this.
Are you still interested?

If I don't hear anything till next Tuesday, I will just merge "as is" and then add my ideas in a new PR.

legacy/setup-debian.sh
laurentaudouin: I integrated your suggestions but I'm completly lost within the interface. I have no idea what to do next. Sorry for being a noob.

laurentaudouin:
Hi Chrisitian,

I thought you'd take it from here in December and since I'm totally 
unfamiliar with github I didn't look forward. This morning I tried to 
implement the meaningful changes you suggested, but I don't know if it's 
done correctly or not in terms of commit and whatnot.

BTW, do you know who could help about R3BRoot ?

Best,

Laurent


Le 14/01/2021 à 23:35, Dr. Christian Tacke a écrit :

legacy/dependencies.md
```diff
@@ -3,12 +3,18 @@
...
 
-* [`legacy/setup-centos-7.sh`](setup-centos-7.sh)
-* [`legacy/setup-debian.sh`](setup-debian.sh)
-* [`legacy/setup-fedora.sh`](setup-fedora.sh)
-* [`legacy/setup-macos.sh`](setup-macos.sh)
-* [`legacy/setup-opensuse.sh`](setup-opensuse.sh)
-* [`legacy/setup-ubuntu.sh`](setup-ubuntu.sh)
+* CentOS 7:
+  [`legacy/setup-centos-7.sh`](setup-centos-7.sh)
+* Debian 10: Please highly consider enabling [packports](https://backports.debian.org/):
```
dennisklein, line 8:
```suggestion
* Debian 10: Please highly consider enabling [backports](https://backports.debian.org/):
```

dennisklein (approved):
small typo, when fixed, pls merge

ChristianTackeGSI:
Thanks @laurentaudouin!

ChristianTackeGSI:
@laurentaudouin wote:
> BTW, do you know who could help about R3BRoot ?

ping: @kresan 

@laurentaudouin : What's up about R3BRoot? In which context are you interested in it? Maybe another communication channel might be more appropriate?

kresan:
@laurentaudouin, you can use https://github.com/R3BRootGroup/R3BRoot to submit a new issue connected to R3BRoot

laurentaudouin:
Hi @kresan thanks for replying, I'm no developper (at least, not yet) ; I've just been trying to install the whole R3BRoot package (I'm in the sofia team). I never managed to do it using "main" versions of FairSoft and FairRoot, which made me kind of crazy. Now I installed older versions of the aforementionned packages and at least R3BRoot compiled. Do you think I did mistakes in my previous attemps, or is there problems in version compatibility ?

ChristianTackeGSI:
Could we move this discussion into a new issue on either FairSoft (because it seems to be FairSoft related) or R3BRoot (because it's R3BRoot related), please?

(it's okay, to mention this PR there, so that the linkage is known.)

ChristianTackeGSI:
Also FairSoft now has a discussion tab here: https://github.com/FairRootGroup/FairSoft/discussions

That might also be a good thing for this type of requests?

laurentaudouin:
Yes, it seems a better way. I'll write there once I have a better view 
of the whole thing.

Thans for your patience, Christian !

L.

Le 24/01/2021 à 18:34, Dr. Christian Tacke a écrit :

kresan:
@laurentaudouin it does not matter whether you are a developer or not. R3BRoot related questions should be discussed in issues on R3BRoot repo. Other R3BRoot users will profit from this.

PR #359, opened 2020-12-16
[spack] root related things

ChristianTackeGSI:
* Add 6.22.x
* cleanup
* improve `meas~llvm` handling in our environments.

### DIFF
--- a/env/dev/sim/spack.yaml
+++ b/env/dev/sim/spack.yaml
@@ -1,10 +1,9 @@
 spack:
   definitions:
   - when: platform == 'linux'
-    root: [root@6.20.08+fortran+gdml+pythia6+pythia8+vc~vdt+python+tmva+mlp+xrootd+sqlite+spectrum
-        ^mesa~llvm]
+    root: [root@6.20.08+fortran+gdml+pythia6+pythia8+vc~vdt+python+tmva+mlp+xrootd+sqlite+spectrum+x+opengl]
   - when: platform == 'darwin'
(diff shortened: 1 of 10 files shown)


dennisklein:
@kresan Can you comment on why we use `mesa~llvm` explicitely? Was it just a concretizer hint or is there some deeper reason for this variant choice?

fuhlig1:
Hi @dennisklein,

maybe I can comment. mesa has some llvm backend which brings in tons of additional dependencies. This slows done the compilation enormously and I haven't seen that it is needed anywhere. So I decided long time ago to switch it off. If I remember correctly I checked at that time and none of the packages from any package manager were providing this back end. This has maybe changed meanwhile and we can debate the previous decision.

kresan:
It was a concretizer hint. ROOT depends on gl and glu. For linux systems mesa is used as a provider for gl. In its turn mesa has a llvm variant, which turns on llvm support and dependency and is on by default. We do not need this part.

dennisklein:
@kresan @fuhlig1 Thx for the context, so I conclude we still want to disable the llvm variant because it causes significant unnecessary compile times.

dennisklein (approved):

ChristianTackeGSI:
> […] If I remember correctly I checked at that time and none of the packages from any package manager were providing this back end. This has maybe changed meanwhile and we can debate the previous decision.

Just for informaation, Debian has it:
```console
root@citpc008:~# aptitude why libllvm7 | grep llvm
i A libgl1-mesa-dri Depends libllvm7 (>= 1:7~svn298832-1~)
```

PR #361, opened 2020-12-23
[legacy] Update dds to 3.5.4

ChristianTackeGSI:


ChristianTackeGSI:
dds 3.5.4 requires C++17.
legacy is compiled with C++14 (I think, due to root?)

So I think, we need to stay at DDS 3.5.3 for a bit.

Comments? (@dennisklein)

ChristianTackeGSI:
marked postponed until we get C++17 in legacy.

ChristianTackeGSI:
@dennisklein is bumping this in his "bump all" PR.

Issue #364, opened 2021-01-05
[legacy] Legacy installation of Nov20 misses gsl


TobiasStockmanns:
In the legacy build of Nov20 the GNU Scientific Library (gsl) is missing. Therefore root::math::MathMore is not generated which is needed in PandaRoot.

ChristianTackeGSI:
I wonder, whether we want to
* add `-Dmathmore=ON` (which is the default) to document PandaRoot's need explicitly.
* add `+math` (which is the default currently) to the spack edition to document PandaRoot's need there as well.
* add `-Dfail-on-missing=ON` so that root's cmake fails early, if things don't match up. (spack has this already.)

karabowi:
I think that adding a flag will not work. We have to add GSL to the list of prerequisites or install it together with legacy.

ChristianTackeGSI:
> I think that adding a flag will not work. We have to add GSL to the list of prerequisites or install it together with legacy.

ACK!

My flag ideas are meant *in addition*.

We have three options (in order of my preference, first being preferred):
1. Add it to the requirements. If we can get it everywhere, I'd vote for that!
   * Debian has `libgsl-dev` which I use in our escape / spack container
   * brew seems to have [this](https://formulae.brew.sh/formula/gsl)
   * Fedora surely has it.
   * opensuse: TBD
2. Adding `-Dbuiltin_gsl=ON` to the ROOT flags
3. Compiling it along with legacy.

dennisklein:
@TobiasStockmanns @karabowi What happens if you install the packages we recommend as system prerequisites: https://github.com/FairRootGroup/FairSoft/blob/nov20/legacy/setup-macos.sh? (Leave out `xrootd` since there is a separate issue with it at the moment: #367)

ChristianTackeGSI:
Ping: @TobiasStockmanns @karabowi 

How should we go on with this?

dennisklein:
> Ping: @TobiasStockmanns @karabowi
> 
> How should we go on with this?

@TobiasStockmanns commented in last meeting that it was likely that he did not have the gsl brew package installed. So it was not really missing, it was just overlooked that we moved it to the system prerequisites.

However, we will add the `-Dmathmore=ON` flag as you proposed and make sure all our `setup-*.sh` scripts list the gsl package. I am on it.

PR #365, opened 2021-01-05
[legacy] root: flags for GSL and XROOTD

ChristianTackeGSI:
* See: #364
  PandaRoot needs mathmore
  * Be explicit about mathmore (more for documentation)
    add `-Dmathmore=ON`

### DIFF
--- a/cmake/legacy.cmake
+++ b/cmake/legacy.cmake
@@ -304,6 +304,7 @@ if(PACKAGE_SET STREQUAL full)
       "-Dglobus=OFF"
       "-Dgnuinstall=ON"
       "-Dhttp=ON"
+      "-Dmathmore=ON"
       "-Dminuit2=ON"
       "-Dmlp=ON"
       "-Dpyroot=ON"


cmake/legacy.cmake
```diff
@@ -295,6 +295,8 @@ if(PACKAGE_SET STREQUAL full)
     URL https://root.cern/download/root_v${root_version}.source.tar.gz
     URL_HASH SHA256=d02f224b4908c814a99648782b927c353d44db79dea2cadea86138c1afc23ae9
     ${CMAKE_DEFAULT_ARGS} CMAKE_ARGS
+      "-Dbuiltin_gsl=ON"
```
dennisklein, line 298: Hm, `gsl` is listed as a prerequisite: https://github.com/FairRootGroup/FairSoft/blob/master/legacy/setup-macos.sh. Does this not work any more?

cmake/legacy.cmake
```diff
@@ -295,6 +295,8 @@ if(PACKAGE_SET STREQUAL full)
...
+      "-Dbuiltin_xrootd=ON"
```
dennisklein, line 299: Before we use this solution I would like to try overwriting the upstream `xrootd` package. Will get back to you this week. In any way, the solution should be conditioned to only be effective on macOS, imho.

ChristianTackeGSI:
Marked "waiting", until we have feedback on #364.

PR #378, opened 2021-02-11
[spack] macos: Fix opengl location

ChristianTackeGSI:
OpenGL is provided by the system.
So let's say it's in /usr.

It's definitely not (any more) in /opt/X11.

### DIFF
--- a/config/darwin/packages.yaml
+++ b/config/darwin/packages.yaml
@@ -1,9 +1,9 @@
 packages:
   opengl:
     paths:
-      opengl@3.3: /opt/X11
-    buildable: False
+      opengl@3.3: /usr
+    buildable: false
   openglu:
     paths:
-      openglu@1.3: /opt/X11
-    buildable: False
+      openglu@1.3: /usr
+    buildable: false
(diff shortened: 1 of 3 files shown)


fuhlig1 (commented):
Actually the libraries are in 

  /usr/local

or better in 

  brew --prefix

ChristianTackeGSI:
Well, there are also some libraires in `/Library/Developer/`?

So I am a bit confused.

```console
alfaci@citmac001 SDKs % pwd && find . -iname glu.h 2>/dev/null
/Library/Developer/CommandLineTools/SDKs
./MacOSX11.1.sdk/System/Library/Frameworks/OpenGL.framework/Versions/A/Headers/glu.h
./MacOSX10.15.sdk/System/Library/Frameworks/OpenGL.framework/Versions/A/Headers/glu.h
```

fuhlig1:
There is something, unfortunately macosx does not bring its X11 installation. That's why we have used the XQuartz DMG package which was installed in /opt/X11. This package installs X11, OpenGL and some other graphics related libraries. 
Since brew now deprecated this package and installs the X11 and OpenGL libraries from source I would use this ones and not the Open GL from the system. 
If we switch off X11 on macosx completely (don't use it in Geant4 and use Cocoa for ROOT) the system version of OpenGL could be a good choice.

ChristianTackeGSI:
Okay, got it!

spack still uses X11 (well, at least it thinks so).
But it builds its own.
So we could let it build its own (mesa based) OpenGL as well.

In the long run, I would prefer to switch off X11 in the spack macos builds as well. Looking into that soon. Just now want to repair the whole spack builds.

ChristianTackeGSI:
@fuhlig1 

This seems to work for now. Can we merge this as is and revisit once we get the X11 deps removed completely from the spack builds?

fuhlig1:
Please merge it. Cleanup can be done later.

fuhlig1 (approved):

PR #385, opened 2021-02-26
C++ standard handling

ChristianTackeGSI:
Add a bunch of "global" (well, environment level) settings of the cxxstd variant. The values resemble what we do in legacy. It is still an open question what to do on a real spack release.

Also a bunch of related stuff. Check the individvual commit messages.


dennisklein (approved):

env/dev/sim/spack.yaml
```diff
@@ -19,6 +19,10 @@ spack:
   concretization: together
   view: false
   packages:
+    all:
+      variants: cxxstd=17
+    xrootd:
+      variants: cxxstd=11
```
dennisklein, line 25: or try xrootd 5?

ChristianTackeGSI: Will do that in the bump versions branch.

Also my intention is to bump spack very soon, which might help in this area as well (xrootd recipe is strange).

ChristianTackeGSI:
@dennisklein can you take a look at the strange workflow build on macos?
https://cdash.gsi.de/testDetails.php?test=7804591&build=263063
I quite don't understand why we can build FairRoot using spack in nearly the same way, but not using the workflow…

I'll merge this now anyhow. FairSoft spack edition isn't in well shape anyway, and this is getting it better somewhat.

dennisklein:
> @dennisklein can you take a look at the strange workflow build on macos?
> https://cdash.gsi.de/testDetails.php?test=7804591&build=263063
> I quite don't understand why we can build FairRoot using spack in nearly the same way, but not using the workflow…
> 
> I'll merge this now anyhow. FairSoft spack edition isn't in well shape anyway, and this is getting it better somewhat.

https://github.com/FairRootGroup/FairMQ/commit/c5487a11ede6e411dafbc9f67c0ef884b4c447d7

PR #398, opened 2021-04-24
[legacy] Add faircmakemodules 0.1.0

dennisklein:


ChristianTackeGSI (approved):
CI looks good.

I wonder, whether we should go for tarball mode with checksum?

dennisklein:
> CI looks good.
> 
> I wonder, whether we should go for tarball mode with checksum?

What's the benefit?

```
/tmp/jenkins-FairRootGroup-FairSoft-PR-398-1.DqD/build/Source/geant4_vmc/.git: No space left on device
```

I will re-run the tests after the other currently running job is through and `/tmp` is cleaned.

ChristianTackeGSI:
> What's the benefit?

I don't know really. For some reason, I like tarball mode.

Maybe:
* checksum protection as a sort of "digitial signature". A git tag can be modified without "our" knowledge.
* Easier caching
* Smaller download

Nothing important, I admit.

PR #401, opened 2021-05-05
[legacy] Update Boost download URL

fuhlig1:
Boost moved from bintray to jfrog.
Update download URL.


ChristianTackeGSI (approved):
Looks good.

PR #403, opened 2021-05-13
[spack] Updated recipes: boost, cmake, dds, r3broot

ChristianTackeGSI:
* Boost 1.76.0
* cmake 3.20.2
* Allow DDS to build with any boost version
* R3BRoot
  * Update patches (will look into upstreaming another time)
  * Add dependency on faircmakemodules
  * Add cxxstd variant to be able to compile with C++17


kresan (approved):
With r3broot recipe is all fine. Thanks.

PR #410, opened 2021-06-09
[legacy] pythia8: Update download URL

### LINKED ISSUE #409: Pythia8 download fails for apr21:legacy
It seems that the download location is not correct anylonger.
The proper webpage should be: https://pythia.org/releases/

dennisklein:
fix #409, will backport to release branches once CI is green.


cmake/legacy.cmake
```diff
@@ -259,10 +259,11 @@ if(PACKAGE_SET STREQUAL full)
 
   list(APPEND packages pythia8)
   set(pythia8_version "8303")
+  set(pythia8_major_version "83")
```
ChristianTackeGSI, line 262:
```suggestion
  string(SUBSTRING "${pythia8_version}" 0 2 pythia8_major_version)
```
What about this?

ChristianTackeGSI (approved):

PR #411, opened 2021-06-09
[spack] pythia8: Update download URL

### LINKED ISSUE #409: Pythia8 download fails for apr21:legacy

dennisklein:
fix #409

### DIFF
--- a/repos/fairsoft/packages/pythia8/package.py
+++ b/repos/fairsoft/packages/pythia8/package.py
@@ -15,6 +15,7 @@ class Pythia8(AutotoolsPackage):
 
     maintainers = ['ChristianTackeGSI']
 
+    version('8305', sha256='98880c3489ba2df0753412d425f3b02e6dfdefb694053b074c48b2af49908779')
     version('8304', sha256='65c6165d61cabefb1594998ff2afebd12616def9bb5d18524212bc5886ca8f68')
     version('8303', sha256='9093351829f92d81c60c719bfb007b1e89efb4307f4c8957407406bf3281a6f7')
     version('8302', sha256='7aae73eabdd38ebbcf7998f63c271ae9b397522497a7ee67a9505b9baae1cb1c')
--- a/repos/fairsoft/packages/pythia8/package.py
+++ b/repos/fairsoft/packages/pythia8/package.py
@@ -11,20 +11,19 @@ class Pythia8(AutotoolsPackage):
        i.e. for the description of collisions at high energies between elementary
        particles such as e+, e-, p and pbar in various combinations."""
 
-    homepage = "http://home.thep.lu.se/Pythia/"
-    url      = "http://home.thep.lu.se/~torbjorn/pythia8/pythia8244.tgz"
+    homepage = "https://pythia.org/"
 
     maintainers = ['ChristianTackeGSI']
 
-    version('8304', sha256='d3897018fb6d545eaf93bf43f32580c984a9bff49259d9dd29dff6edfbe9d9a1')
-    version('8303', sha256='cd7c2b102670dae74aa37053657b4f068396988ef7da58fd3c318c84dc37913e')
-    version('8302', sha256='7372e4cc6f48a074e6b7bc426b040f218ec4a64b0a55e89da6af56933b5f5085')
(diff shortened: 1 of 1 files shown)


repos/fairsoft/packages/pythia8/package.py
```diff
@@ -11,27 +11,21 @@ class Pythia8(AutotoolsPackage):
...
 
-    version('8304', sha256='d3897018fb6d545eaf93bf43f32580c984a9bff49259d9dd29dff6edfbe9d9a1')
-    version('8303', sha256='cd7c2b102670dae74aa37053657b4f068396988ef7da58fd3c318c84dc37913e')
-    version('8302', sha256='7372e4cc6f48a074e6b7bc426b040f218ec4a64b0a55e89da6af56933b5f5085')
-    version('8301', sha256='51382768eb9aafb97870dca1909516422297b64ef6a6b94659259b3e4afa7f06')
-    version('8244', sha256='e34880f999daf19cdd893a187123927ba77d1bf851e30f6ea9ec89591f4c92ca')
-    version('8240', sha256='d27495d8ca7707d846f8c026ab695123c7c78c7860f04e2c002e483080418d8d')
-    version('8235', sha256='e82f0d6165a8250a92e6aa62fb53201044d8d853add2fdad6d3719b28f7e8e9d')
-    version('8230', sha256='332fad0ed4f12e6e0cb5755df0ae175329bc16bfaa2ae472d00994ecc99cd78d')
-    version('8212', sha256='f8fb4341c7e8a8be3347eb26b00329a388ccf925313cfbdba655a08d7fd5a70e')
```
ChristianTackeGSI, line 27: Please keep the old versions.
The recipe is slowly converging with spack upstream and old version don't hurt.
Or do they hurt?

repos/fairsoft/packages/pythia8/package.py
```diff
@@ -11,27 +11,21 @@ class Pythia8(AutotoolsPackage):
...
+    version('8305', sha256='98880c3489ba2df0753412d425f3b02e6dfdefb694053b074c48b2af49908779')
+    version('8304', sha256='65c6165d61cabefb1594998ff2afebd12616def9bb5d18524212bc5886ca8f68')
+    version('8303', sha256='9093351829f92d81c60c719bfb007b1e89efb4307f4c8957407406bf3281a6f7')
+    version('8240', sha256='3d71cae2b6fc451910b69a18ee4f6d527c5ad10e32982a0dce5f6449a10c3ff1')
```
ChristianTackeGSI, line 21: Wow, even the checksums changed. Did you have a look, why they changed?

repos/fairsoft/packages/pythia8/package.py
```diff
@@ -49,3 +43,7 @@ def configure_args(self):
     def setup_environment(self, spack_env, run_env):
         run_env.set('PYTHIA8DATA', '%s/share/Pythia8/xmldoc' % prefix )
         run_env.set('PYTHIA8', '%s' % prefix )
+
+    def url_for_version(self, version):
+        url = "https://pythia.org/download/pythia{0}/pythia{1}.tgz"
+        return url.format(str(version)[:2], version)
```
ChristianTackeGSI, line 49: Usually, spack's url-version-magic can handle two entries of the version code, even with subgroups and things.
Naive question: Did you try to put an `url` keyword and it didn't work out?
Hmm. Strange.

ChristianTackeGSI: After having seen this in legacy again, I realized, that for spack this is only one single version part. Not like `83.03`. So it can't see `83` as a part of `8303`.

Resolving.

repos/fairsoft/packages/pythia8/package.py
dennisklein: Have not checked why they changed the checksums when moving to the new website. That is also the reason why I was too lazy to look all of them up. Upstream has not yet adapted, so there was nothing to copy from :P

repos/fairsoft/packages/pythia8/package.py
dennisklein: Added all previous versions and split commit into two.

ChristianTackeGSI (approved):

PR #412, opened 2021-06-10
[legacy] Multiple updates

dennisklein:
+ standalone `asio` package in preparation of upcoming `FairMQ` and `asiofi` releases


fuhlig1 (approved):

dennisklein:
@fuhlig1 Should we also bump https://github.com/vmc-project/vgm/commits/v4-9 and https://pythia.org/history/ to latest?

dennisklein:
The ODC bump does not build in this PR, but @AndreyLebedev is working on a fix directly in ODC, so switching this PR to draft until this is finished, so no package-level patch is required.

AndreyLebedev:
@dennisklein The patch is in master now. Do you also need a new tag for ODC?

dennisklein:
> @dennisklein The patch is in master now. Do you also need a new tag for ODC?

Thx, yes, tag would be good, but let's wait (I can test without a tag and only once everything looks good, it is worth tagging from my side. Of course, if you have other reasons in the meantime for a tag, don't wait on me).

AndreyLebedev:
DDS is fixed now. The tag is the same.

AndreyLebedev:
ODC is now `0.26`

PR #420, opened 2021-07-23
pass on CMAKE_TOOLCHAIN_FILE, if present

klenze:
If the user specifies a toolchain file, pass it on to the various cmake calls used to configure the individual programs. 

This allows us to overcome cmake's natural +5 resistance to environment variables and make it use a different compiler than e.g. /usr/bin/c++ for all of FairSoft, which is a great boon on systems where the system compiler still comes on punch cards.


dennisklein (approved):

dennisklein:
Thx! It will be included in the `jul21` release which is landing within the next days. If you need backports to older releases, let me know.

PR #422, opened 2021-08-02
[spack] Improve for Fedora-34; Bump dds,odc

ChristianTackeGSI:
* Some CI changes to better support Fedora-34
* Bump dds with 3.5.16, odc 0.34
* Bump xrootd to 5.3.1.
* Backport xrootd5 support to ROOT 6.22.08.


ChristianTackeGSI:
Note:

Current problem is xrootd 4 on Fedora-34 (gcc11).

We could try to port
 https://github.com/FairRootGroup/FairSoft/commit/db9ca700789f0667e460e00e74940961db1009d1
to spack and thus allow older ROOT versions to build with xrootd 5.
I am unsure, whether it is worth doing that?

test/workflow/classic_developer.sh
```diff
@@ -18,7 +18,7 @@ fairroot_branch=dev
 case "$ENVNAME" in
   *jun19*)
     fairroot_branch=v18.6_patches
-    no_boost_cmake=1
+    extra_cmake_flags="-DBoost_NO_BOOST_CMAKE=ON -DUSE_DIFFERENT_COMPILER=ON"
```
dennisklein, line 21: Why is it needed?

dennisklein:
> Note:
> 
> Current problem is xrootd 4 on Fedora-34 (gcc11).
> 
> We could try to port
> [db9ca70](https://github.com/FairRootGroup/FairSoft/commit/db9ca700789f0667e460e00e74940961db1009d1)
> to spack and thus allow older ROOT versions to build with xrootd 5.
> I am unsure, whether it is worth doing that?

I guess, if those patches just work, then let's use them, if not, then what options do we have? disable xrootd in root on `gcc11 && nov20`?

dennisklein:
FYI: Mattias Ellert is the maintainer of the root package in fedora and epel.

test/workflow/classic_developer.sh
ChristianTackeGSI: TBH, I have just refactored this. I haven't looked deeply enough, wether `USE_DIFFERENT_COMPILER` or the boost flag are actually needed for 18.6.

I just realized, that FairRoot/dev doesn't need the `USE_DIFFERENT_COMPILER` one.

dennisklein: I remember adding support for homebrew and ccache compiler wrappers in v18.6.0 (e21d430bf5cb7a6e2f7b1fc0d70523d0f8f81cf5). Not sure about spack though.

ChristianTackeGSI:
> I guess, if those patches just work, then let's use them, if not, then what options do we have? disable xrootd in root on `gcc11 && nov20`?

They work, including upgrading xrootd.

So jun19 is now disabled on Fedora-34.

We could try disabling xrootd on jun19? But I don't know, if it is worth it?

test/workflow/classic_developer.sh
ChristianTackeGSI: One possible way of thinking here might be:
The spack world is compiled with the spack wrapper.
But the workflow uses the normal compiler (and the user is responsible for choosing matching flags).
So compiler match checking isn't really easy.

ChristianTackeGSI: Can you please propose something to do next regarding this?

dennisklein: I don't think there is anything to do.

dennisklein (approved):

PR #425, opened 2021-08-14
CI Update for Slurm; Add Debian-11

ChristianTackeGSI:
Our nodes are heterogeneous.
So allocate a whole node and use it fully.

Allocate from the "long" partition, so that scheduling can reserve some space for shorter jobs.

Changed the whole thing over to not use `--hint=compute_bound` and thus calculate everything in units of threads, not cores.
* Our macs are confugured with `SLURM_CPUS_PER_TASK=12`
* CMake's `ProcessorCount()` gives threads


ChristianTackeGSI:
This is part of the aug21rc branch. Will hopefully find its way into dev soon.

PR #429, opened 2021-09-08
[legacy] Add libreadline-dev to Ubuntu setup

rbx:
The package seems to be needed for Ubuntu 21.04. Other setups, such as centos7 include it.


ChristianTackeGSI:
Looks good!

Thanks!

(I deleted your merge branch, hope that's fine with you.)

We probably should sync this into `setup-debian.sh`, even if it's not needed there yet.

PR #434, opened 2021-10-25
[spack] root: +fftw; Add private PandaRoot recipe

ChristianTackeGSI:
PandaRoot will need the fftw feature in ROOT. So add it to fairsoft-bundle
See: #445

Also add a private pandaroot recipe and environment for internal testing purposes.

Maybe one day we will add something like this to our CI. So that our CI tests whether any changes to FairSoft break PandaRoot.

The current environment builds PandaRoot/dev against FairRoot/dev and FairSoft/dev.

---

In addition this PR contains the current setup needed to install the CI images on our CD machines.

It also starts to add libfftw3-dev to the debian/ubuntu legacy setups. (the containers on cvmfs already have that).

### DIFF
--- a/test/container/CMakeLists.txt
+++ b/test/container/CMakeLists.txt
@@ -6,6 +6,15 @@
 #                  copied verbatim in the file "LICENSE"                       #
 ################################################################################
 
+include(GNUInstallDirs)
+
+if(NOT ${PROJECT_NAME}_CONTAINER_DIR)
+  set(${PROJECT_NAME}_CONTAINER_DIR "${CMAKE_INSTALL_DATADIR}/${PROJECT_NAME_LOWER}/container"
+      CACHE PATH "CI Containers (DATADIR/${PROJECT_NAME_LOWER}/container)")
+endif()
+mark_as_advanced(${PROJECT_NAME}_CONTAINER_DIR)
+set(PROJECT_CONTAINER_DIR "${${PROJECT_NAME}_CONTAINER_DIR}")
+
 add_subdirectory(legacy)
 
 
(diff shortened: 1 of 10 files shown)


test/container/CMakeLists.txt
```diff
@@ -6,6 +6,12 @@
 #                  copied verbatim in the file "LICENSE"                       #
 ################################################################################
 
+include(GNUInstallDirs)
+
+if(NOT DEFINED FairSoft_CONTAINER_DIR)
+  set(FairSoft_CONTAINER_DIR "${CMAKE_INSTALL_DATADIR}/${PROJECT_NAME}")
```
dennisklein: 1. In FairMQ, ODC, etc I named this kind of "project-specific install tree path-component variables" `PROJECT_<component>[_]DIR`. In the very early times I had the concrete project name in there, but then realized it is always the same pattern and in one case was even annoying because I changed the project name capitalization and had to change now all those variables to keep it consistent. If it did not cross your mind before, maybe you want to consider, if `PROJECT_CONTAINER_DIR` would be suitable here, too. But I see a small but significant? difference to the other repos: In this case you allow the user to customize it, in the other repos the user can only customize the `CMAKE_INSTALL_<component>[_]DIR` part.

2.
```suggestion
  set(FairSoft_CONTAINER_DIR "${CMAKE_INSTALL_DATADIR}/${PROJECT_NAME_LOWER}")
```
If we do not have `PROJECT_NAME_LOWER` yet, please define it in the main CMakeLists.txt somewhere near and after the `project()` command (`STRING(TO_LOWER ...` or similar was the command).

dennisklein (changes requested):

test/container/CMakeLists.txt
ChristianTackeGSI: Both should be addressed now.

ChristianTackeGSI:
Just again, ubuntu works at least halfway: https://cdash.gsi.de/viewTest.php?buildid=315944

dennisklein (approved):

PR #448, opened 2022-02-22
Bump Geant4 to 11.0.0

dennisklein:
fix #447

---

Just testing this against apr21_patches, will change it to dev branch, as soon as I found time to update it.

PR #451, opened 2022-02-23
cherry-pick / port multiple things from apr21_patches

ChristianTackeGSI:
I basically took apr21_patches and rebased it (in slow steps) on top of dev. Then cleaned it up a bit.

Open Issues:
* Some commits could need some rewording maybe (currently they're just the plain commit texts from apr21_patches)
* dds: In dev the cmake arg `-DBoost_NO_BOOST_CMAKE=ON` was removed, it apr21_patches it still exists. What's the story? What do we want?
* rebuilding the containers on our CD machines for this. (Christian will do soon.)


dennisklein:
awesome, thx! could you also pick the bumps in https://github.com/FairRootGroup/FairSoft/pull/448 (caveat: there is a problem with boost python still, but the bumps in principle will be needed anyways in dev)

AnarManafov:
@dennisklein , I definetly can make this option disabled by default. It's a bootstrap script, so nothing is set in stone in there. I think, it worked for some time without this options, but with recent tests on boost 1.76 it started to fail again. This is why I enabled it back.

I also wanted to bump minimum cmake and boost versions in DDS.
What minimums do you use (want to use in the new future) in FairSoft?

dennisklein:
> @dennisklein , I definetly can make this option disabled by default. It's a bootstrap script, so nothing is set in stone in there. I think, it worked for some time without this options, but with recent tests on boost 1.76 it started to fail again. This is why I enabled it back.

To me, I don't really have a preference, @ChristianTackeGSI was asking whether we need it and I thought maybe you have more context info than I had.

@ChristianTackeGSI Regarding the flag in question here, I would probably just follow the default install instructions of DDS.

> I also wanted to bump minimum cmake and boost versions in DDS. What minimums do you use (want to use in the new future) in FairSoft?

DDS 3.6 is part of the FairSoft `apr21p2` release which ships with Boost 1.75. Next major release we are now preparing (likely `feb22` or if we are too slow, then `mar22`) and it will have Boost 1.78, if we do not encounter any showstoppers (only just bumped it and started testing).

However, I would not set this newest Boost version as requirement in DDS, only, if DDS really depends on APIs/features in this version. I recommend to **require** the lowest possible version only. But its of course up to you.

ChristianTackeGSI:
Update:

* I have rebuild all containers on our CD machines from this branch. They're in `/cvmfs/fairsoft_dev.gsi.de/ci/for-fairsoft/2022-02-23_1829/container/`. I have symlinked them all into the CI area.
* I have re-added the `-DBoost_NO_BOOST_CMAKE=ON` flag to the DDS build
* I have rebased `dklein/apr21_patches` from #448 on top of this branch.

If anything doesn't match, please let me know, then we'll fix it.

dennisklein (approved):

PR #461, opened 2022-03-29
[legacy] Fix dependency tracking for pkgs with `URL` source

### LINKED ISSUE #460: Unable to modify / rebuild ROOT.
The ROOT package is not rebuild after modifying the source. I.e. changing a file under `build/Source/root` and running `cmake --build build/ --target root` only outputs:
```
[ 20%] Built target hepmc
[ 40%] Built target pythia6
[ 60%] Built target vc
[ 80%] Built target pythia8
[100%] Built target root
```
Not updating the existing installation.

dennisklein:
fix #460

### DIFF
--- a/cmake/legacy.cmake
+++ b/cmake/legacy.cmake
@@ -134,6 +134,7 @@ ExternalProject_Add(boost
   URL "https://boostorg.jfrog.io/artifactory/main/release/1.${boost_version}.0/source/boost_1_${boost_version}_0.tar.bz2"
   URL_HASH SHA256=8681f175d4bdb26c52222665793eef08490d7758529330f98d3b29dd0735bccc
   BUILD_IN_SOURCE ON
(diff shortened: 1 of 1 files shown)


ChristianTackeGSI:
release tarballs are a good thing:

* They can be "signed" by us using the `URL_HASH` method. So that people actually get the exact same source always.
* If a release tarball exists, it is usually the "official source" from which projects expect users to build their software from.

dennisklein:
> release tarballs are a good thing:

well, very debatable statement ;)

ChristianTackeGSI:
> well, very debatable statement ;)

Okay, added a "IMHO". ;-)

fuhlig1:
I think the change is fine with me, as it solves the issue. In my opinion we can tolerate the overhead with running unneeded reinstallation. This feature is probably not used by many users.

PR #462, opened 2022-03-30
[legacy] release preps

dennisklein:


cmake/legacy.cmake
```diff
@@ -211,29 +211,26 @@ ExternalProject_Add(flatbuffers
...
+  # ExternalProject_Add(odc
+    # GIT_REPOSITORY https://github.com/FairRootGroup/ODC GIT_TAG ${odc_version}
+    # ${CMAKE_DEFAULT_ARGS} CMAKE_ARGS
+      # "-DBUILD_GRPC_CLIENT=OFF"
+      # "-DBUILD_GRPC_SERVER=OFF"
+      # "-DBUILD_EPN_PLUGIN=OFF"
+      # "-DBUILD_EXAMPLES=OFF"
+    # DEPENDS boost dds fairlogger fairmq ${extract_source_cache_target}
+    # ${LOG_TO_FILE}
+  # )
```
ChristianTackeGSI, line 233: What about moving the odc-setup into some `add_package_odc` macro and just not call it?

ChristianTackeGSI (approved):

cmake/legacy.cmake
dennisklein: ya, I should probably just have deleted the section. It's just noise. Will do on next PR.

PR #463, opened 2022-04-08
More release prep

dennisklein:


ChristianTackeGSI (approved):
My feeling about a ROOT .00 release isn't perfect. But I have no hard reasons not to switch to 6.26. So it's okay for me.

dennisklein:
> My feeling about a ROOT .00 release isn't perfect. But I have no hard reasons not to switch to 6.26. So it's okay for me.

I had a similar concern at first, but then I saw that it is clearly marked as production release. Also, I think we can relax the *physics-sensitive-package-version-pin* to major/minor versions only in the case of Root.

PR #464, opened 2022-04-25
Fix compiler error on macosx

fuhlig1:
At least on macosx 10.5 with only the command line tools installed the zeromq
compilation crashes because the gnutls headers are not found. Those headers
are on this system not installed in a system path but properly found by CMake.
Unfortunately the include directory wasn't added when compiling the sources.

### DIFF
--- a/cmake/legacy.cmake
+++ b/cmake/legacy.cmake
@@ -201,6 +201,7 @@ ExternalProject_Add(zeromq
     "-DWITH_PERF_TOOL=ON"
     "-DZMQ_BUILD_TESTS=ON"
     "-DENABLE_CPACK=OFF"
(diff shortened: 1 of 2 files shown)


dennisklein:
hm, we could also just add the configure option `-DENABLE_DRAFTS=OFF` in `cmake/legacy.cmake`. What do you think?

fuhlig1:
If we don't need it we can switch it off. You know much better which zeromq features are needed for FairMQ.

dennisklein:
> If we don't need it we can switch it off. You know much better which zeromq features are needed for FairMQ.

I have added a patch to #465

fuhlig1:
@dennisklein,

ich will test after the meetings.

dennisklein:
fixed by #465

PR #465, opened 2022-04-28
GCC12 support and switching off experimental zmq features

dennisklein:


cmake/legacy.cmake
```diff
@@ -199,6 +199,7 @@ ExternalProject_Add(zeromq
     "-DWITH_PERF_TOOL=ON"
     "-DZMQ_BUILD_TESTS=ON"
     "-DENABLE_CPACK=OFF"
+    "-DENABLE_DRAFTS=OFF"
```
fuhlig1, line 202: The change worked for me.

fuhlig1, line 202: The change worked for me.

ChristianTackeGSI (approved):

PR #466, opened 2022-04-29
Hint `ICU_ROOT/lib` to DDS

dennisklein:
@fuhlig1 Let me know, if this resolves your build issue with the `wn_bin` target of DDS.

### DIFF
--- a/cmake/legacy.cmake
+++ b/cmake/legacy.cmake
@@ -165,12 +165,16 @@ ExternalProject_Add(fmt
   ${DEPENDS_ON_SOURCE_CACHE}
 )
 
+if(ICU_ROOT)
+  set(dds_icu_hint "-DDDS_LD_LIBRARY_PATH=${ICU_ROOT}/lib")
+endif()
 list(APPEND packages dds)
 set(dds_version "3.6")
 ExternalProject_Add(dds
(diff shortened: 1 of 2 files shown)


fuhlig1 (approved):
Works for me.

ChristianTackeGSI:
Two questions:

1. Should we try to get this patch merged into DDS upstream?

2. Shizld we add that patch to the spack packaging? (desipute not being needed there)

karabowi:
On my mac (11.6) I had the same ICU problem. This patch fixed the DDS build for me as well.

fuhlig1:
we should propose the patch to Anar.

dennisklein:
If anyone is making PRs, consider to propose both patches from https://github.com/FairRootGroup/FairSoft/tree/dev/legacy/dds

ChristianTackeGSI:
* `fix_boost_lookup.patch` – looks like it already is in this commit: 
  https://github.com/FairRootGroup/DDS/commit/8c6c950ed013da570da2120f03c484cb62fb6014
* `allow_dds_ld_library_path_as_cmake_var.patch` – PR here:
  https://github.com/FairRootGroup/DDS/pull/438

PR #486, opened 2022-11-18
[legacy] onnxruntime: Fix compilation error

### LINKED ISSUE #485: Building of ONNXRuntime crashes
A problem when building ONNXRuntime was reported by a CBM user a week ago. Up to now I wasn't able to reproduce the problem but today when building the nov22rc prerelease in our Debian10 VAE I ran in the same problem. I attach the original mail, below.

`-    onnxruntime (oder externals davon) benötigt cmake >= 3.18 (FairRoot nur 3.16...)`

This issue s fixed in the release candidate branch

```
-   *.proto files werden in '**/FairSoft/build/Build/onnxruntime/external/onnx/onnx/' erwartet, sind aber nicht da. (siehe error_proto.txt)
    Mein quick fix war die files von '**/FairSoft/build/Source/onnxruntime/cmake/external/onnx/onnx/' zu kopieren ...
-   Beim onnx konfigurieren, wodurch in cmake teilweise python2 verwendet wird, was beim builden zu einem syntax error führt. (sieht error_python2.txt)
    Gelöst durch 'https://github.com/onnx/onnx/issues/4469#issuecomment-1260274285' .
    In file '**/FairSoft/build/Source/onnxruntime/cmake/external/onnx/CMakeLists.txt' line 116-118
```

This are the problems I ran into today. In the Debian10 Container we have Python2 and Python3 installations such that generation of some protobuffer input files crashes with an error message (second problem from above).  The fix propose to uncomment the first line of the code below and comment line 2 and 3.

```
#find_package(Python ${PY_VERSION} COMPONENTS Interpreter Development REQUIRED)
find_package(PythonInterp ${PY_VERSION} REQUIRED)
find_package(PythonLibs ${PY_VERSION})

```

When applying the fix proposed in the link above the installation stops with the first error from above. Actually there are really some *.proto files in the source directory but these are not produced but are already in the repository. So the fix done by Martin wasn't correct at least in my opinion. Looking more closely to the build system code I realised that the *.proto files are produced using a python script which is executed via add_custom_command() using the CMake variable PYTHON_EXECUTABLE. After using   
`find_package(Python` the name of the variable is unfortunately Python_EXECUTABLE. After changing also this in the CMakeLists.txt the code compiles without problems.

```
-   libonnxruntime.so ist bei mir 731 MB, vermutlich zu groß ? Dabei wurde die RelWithDebInfo version verwendet, 
    Release sollte drastisch kleiner sein ( < 20 MB ?).
```

This is probably due to the used WSL2 (Windows-Subsystem for Linux). Martin at least did not report problems when using ARCH Linux.

I will create a pull request with the fix and propose to add the fix in the build system to our existing patch. I think this will not introduce new problems but that we should see with the CI.

fuhlig1:
In case a mixed Python2 and Python3 installation is on the system the needed files for protobuffer can't be created and the compilation stops with an error. The patch fixes the problem by properly finding and using the correct Python installation.
Fixes #485 .

### DIFF
--- a/cmake/legacy.cmake
+++ b/cmake/legacy.cmake
@@ -430,6 +430,7 @@ list(APPEND packages onnxruntime)
 set(onnxruntime_version "1.12.1")
 ExternalProject_Add(onnxruntime
   PATCH_COMMAND ${patch} -p1 -i "${CMAKE_SOURCE_DIR}/legacy/onnxruntime/install_config_files.patch"
+  COMMAND ${patch} -p1 -i "${CMAKE_SOURCE_DIR}/legacy/onnxruntime/fix_python_detection.patch"
   GIT_REPOSITORY https://github.com/microsoft/onnxruntime/ GIT_TAG v${onnxruntime_version}
   GIT_SHALLOW ON
   GIT_SUBMODULES
--- a/legacy/onnxruntime/fix_python_detection.patch
+++ b/legacy/onnxruntime/fix_python_detection.patch
@@ -1,0 +1,26 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index f914e78b..39e22923 100644
+--- a/cmake/external/onnx/CMakeLists.txt
++++ b/cmake/external/onnx/CMakeLists.txt
+@@ -113,9 +113,9 @@ endif()
+ 
+ # find_package Python has replaced PythonInterp and PythonLibs since cmake 3.12
+ # Use the following command in the future; now this is only compatible with the latest pybind11
+-# find_package(Python ${PY_VERSION} COMPONENTS Interpreter Development REQUIRED)
+-find_package(PythonInterp ${PY_VERSION} REQUIRED)
+-find_package(PythonLibs ${PY_VERSION})
++ find_package(Python ${PY_VERSION} COMPONENTS Interpreter Development REQUIRED)
++#find_package(PythonInterp ${PY_VERSION} REQUIRED)
++#find_package(PythonLibs ${PY_VERSION})
+ 
+ if(CMAKE_SYSTEM_NAME STREQUAL "AIX")
+   set(CMAKE_NO_SYSTEM_FROM_IMPORTED 1)
+@@ -256,7 +256,7 @@ function(RELATIVE_PROTOBUF_GENERATE_CPP NAME SRCS HDRS ROOT_DIR DEPEND)
+     endif()
+ 
+     add_custom_command(OUTPUT "${GENERATED_PROTO}"
+-                       COMMAND "${PYTHON_EXECUTABLE}" "${GEN_PROTO_PY}"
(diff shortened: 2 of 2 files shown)


fuhlig1:
@dennisklein,

did you also test the compilation of the onnxruntime target or only the default target?

dennisklein:
> @dennisklein,
> 
> did you also test the compilation of the onnxruntime target or only the default target?

See f3a11ee25286a8d5da075bf39da468518155071e, it only builds the target `onnxruntime`, but there are no additional plausibility checks in place at the moment.

fuhlig1:
> > @dennisklein,
> > did you also test the compilation of the onnxruntime target or only the default target?
> 
> See [f3a11ee](https://github.com/FairRootGroup/FairSoft/commit/f3a11ee25286a8d5da075bf39da468518155071e), it only builds the target `onnxruntime`, but there are no additional plausibility checks in place at the moment.

Thanks I overlooked that. For my concern it is okay that there are no further checks available yet. I was only interested If my change introduces any problems in our test systems. 

According to the output of the tests this isn't the case. The failing test fails due to some other problem, the compilation of onnxruntime worked.

legacy/onnxruntime/install_config_files.patch
```diff
@@ -115,3 +115,29 @@ index 700a0e264..e659b1390 100644
...
+   set(CMAKE_NO_SYSTEM_FROM_IMPORTED 1)
+@@ -256,7 +256,7 @@ function(RELATIVE_PROTOBUF_GENERATE_CPP NAME SRCS HDRS ROOT_DIR DEPEND)
+     endif()
+ 
+     add_custom_command(OUTPUT "${GENERATED_PROTO}"
+-                       COMMAND "${PYTHON_EXECUTABLE}" "${GEN_PROTO_PY}"
++                       COMMAND "${Python_EXECUTABLE}" "${GEN_PROTO_PY}"
+                                ARGS ${GEN_PROTO_ARGS}
+                        DEPENDS ${INFILE}
+                        COMMENT "Running gen_proto.py on ${INFILE}"
```
dennisklein, line 143: Please move this to a different patch file, e.g. `legacy/onnxruntime/fix_python_detection.patch` or similar.

dennisklein (changes requested):
Please adopt the message in the patch file name for your commit message header, e.g. instead of `[legacy] onnxruntime: Fix compilation error` use `[legacy] onnxruntime: Fix Python detection`

legacy/onnxruntime/install_config_files.patch
fuhlig1: I wanted to do that but I am not sure how do I call the second patch from our CMake.

fuhlig1: Okay. Now I found it. I will do as done for dds.

fuhlig1: Done.

dennisklein (approved):

PR #488, opened 2022-12-06
[legacy] root: Fix compilation for CMake >= 3.24

### LINKED ISSUE #487: [nov22]: ROOT compilation fails on Debian11
With the latest version of Debian11 I the build of ROOT fails with the error below. The compilation was working last week and since I don't see any changes on our side it is probably related to a change of the OS itself.

```
gmake[4]: *** [Makefile:156: all] Error 2

CMake Error at /opt/fairsoft/source/FairSoft/build/Stamp/root/root-build-RelWithDebInfo.cmake:47 (message):
  Stopping after outputting logs.


gmake[3]: *** [CMakeFiles/root.dir/build.make:85: Stamp/root/root-build] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:3450: CMakeFiles/root.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:3457: CMakeFiles/root.dir/rule] Error 2
gmake: *** [Makefile:1694: root] Error 2
```

Checking in more detail it turns out that the compilation of LZMA fails

```
Singularity> cd build/Build/root/
Singularity> make
[  0%] Performing build step for 'LZMA'
CMake Error at /opt/fairsoft/source/FairSoft/build/Build/root/LZMA-prefix/src/LZMA-stamp/LZMA-build-RelWithDebInfo.cmake:49 (message):
  Command failed: 2

   'make'

  See also

    /opt/fairsoft/source/FairSoft/build/Build/root/LZMA-prefix/src/LZMA-stamp/LZMA-build-*.log


make[2]: *** [CMakeFiles/LZMA.dir/build.make:86: LZMA-prefix/src/LZMA-stamp/LZMA-build] Error 1
make[1]: *** [CMakeFiles/Makefile2:7674: CMakeFiles/LZMA.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```
 
with the following underlying error

```
Singularity> more /opt/fairsoft/source/FairSoft/build/Build/root/LZMA-prefix/src/LZMA-stamp/LZMA-build-*.log
::::::::::::::
/opt/fairsoft/source/FairSoft/build/Build/root/LZMA-prefix/src/LZMA-stamp/LZMA-build-err.log
::::::::::::::
/opt/fairsoft/source/FairSoft/build/Build/root/LZMA-prefix/src/LZMA/build-aux/missing: line 81: aclocal-1.15: command not found
WARNING: 'aclocal-1.15' is missing on your system.
         You should only need it if you modified 'acinclude.m4' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'aclocal' program is part of the GNU Automake package:
         <http://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <http://www.gnu.org/software/autoconf>
         <http://www.gnu.org/software/m4/>
         <http://www.perl.org/>
make[3]: *** [Makefile:514: aclocal.m4] Error 127
```

which is probably correct since aclocal 1.16 is installed on the system. 

I was not able to fix the installation error so I simply installed liblzma on the system such that ROOT could use that version and don't need to compile its internal version. With this change the problem disappears.

I will create a PR which adds the missing package to legacy/setup-debian.sh.

fuhlig1:
This PR fixes #487.

### DIFF
--- a/cmake/legacy.cmake
+++ b/cmake/legacy.cmake
@@ -350,6 +350,7 @@ ExternalProject_Add(root
   COMMAND ${patch} -p1 -i "${CMAKE_SOURCE_DIR}/legacy/root/fix_pyroot_code_h_inclusion_python_3_11.patch"
   COMMAND ${patch} -p1 -i "${CMAKE_SOURCE_DIR}/legacy/root/fix_tpython_python_3_11_deprecation.patch"
   COMMAND ${patch} -p1 -i "${CMAKE_SOURCE_DIR}/legacy/root/fix_pyroot_cast_error_python_3_11.patch"
+  COMMAND ${patch} -p1 -i "${CMAKE_SOURCE_DIR}/legacy/root/fix_root_install_external_tars_with_cmake_3_24.patch"
   ${CMAKE_DEFAULT_ARGS} CMAKE_ARGS
     "-Daqua=ON"
     "-Dasimage=ON"
--- a/legacy/root/fix_root_install_external_tars_with_cmake_3_24.patch
+++ b/legacy/root/fix_root_install_external_tars_with_cmake_3_24.patch
@@ -1,0 +1,96 @@
+commit e79ef2e76ea8e20c9e4edc28844b97f82122033f
+Author: Axel Naumann <Axel.Naumann@cern.ch>
+Date:   Fri Nov 18 15:32:34 2022 +0100
+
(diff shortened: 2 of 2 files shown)


fuhlig1:
@dennisklein, 

how do you get the change into nov22_patches branch? Do you use cherry picking?

PR #492, opened 2023-01-02
[legacy] patches for compiling dds, clhep and pythia8 with c++20.

klenze:
These are not yet applied in legacy.cmake.

When trying to compile FairSoft with C++20, I had to change a few things:
* clhep manually checks for version strings and refuses c++20 in cmake. If that check it bypassed, it compiles. 
* dds and pythia8 sometimes put template brackets in the names of constructors and destructors, which C++20 [forbids](https://eel.is/c++draft/diff.cpp17#class). 

Both of these were easy enough to fix. Then I tried to compile root with c++20, which went poorly. TL;DR: cling does not include several C++20 constructs like ``std::construct_at``, which libstdc++ headers assume to be available if ``__cplusplus > 201703``. I tried making cling set __cplusplus to 201703, but that failed for some other reason, so for now I will just keep ROOT on C++17 and try to run everything else with C++20 and hope that for the best (e.g. no repetition of _GLIBCXX_USE_CXX11_ABI).

Anyhow, I am sure that C++20 support will arrive in root Really Soon Now and then these patches might be helpful to fully support C++20.

### DIFF
--- a/legacy/clhep/clhep_cxx20.patch
+++ b/legacy/clhep/clhep_cxx20.patch
@@ -1,0 +1,13 @@
+diff --git a/cmake/Modules/ClhepVariables.cmake b/cmake/Modules/ClhepVariables.cmake
+index ae48462..1f85cc0 100644
+--- a/cmake/Modules/ClhepVariables.cmake
++++ b/cmake/Modules/ClhepVariables.cmake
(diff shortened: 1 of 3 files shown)


dennisklein:
dev now compiles with C++20

PR #508, opened 2023-12-18
[legacy] `jan24` RC

dennisklein:


cmake/legacy.cmake
```diff
@@ -306,12 +296,12 @@ else()
...
 ExternalProject_Add(geant4
   URL https://geant4-data.web.cern.ch/releases/geant4-v${geant4_version}.tar.gz
-  URL_HASH SHA256=72f8b687512d6e630dc6c64c5c41ba711220468540132999d00e4fce1908ae48
+  URL_HASH SHA256=46ad7fab3c5cb4bd0bdd77dd6d3e2283184819235bcbc01b2d117d81b35596a6
   ${CMAKE_DEFAULT_ARGS} CMAKE_ARGS
     "-DCMAKE_CXX_STANDARD=${CMAKE_CXX_STANDARD}"
     ${mt}
     "-DGEANT4_USE_SYSTEM_CLHEP=ON"
-    "-DGEANT4_USE_SYSTEM_EXPAT=ON"
+    "-DGEANT4_USE_SYSTEM_EXPAT=OFF"
```
ChristianTackeGSI, line 304: I don't think, this is needed.

We can stay with `GEANT4_USE_SYSTEM_EXPAT=ON`.

dennisklein: @MohammadAlTurany Can you comment why this was needed (took it from your branch)?

dennisklein: @MohammadAlTurany Can you share the output of `brew info expat`?

We don't list `expat` explicitely in [`setup-macos.sh`](/legacy/setup-macos.sh), but we have `mesa-glu` which depends on `expat` (see `brew deps mesa-glu`). So, technically, the `expat` brew package should be installed.

MohammadAlTurany: `
==> expat: stable 2.5.0 (bottled), HEAD [keg-only]
XML 1.0 parser
https://libexpat.github.io/
/usr/local/Cellar/expat/2.5.0 (21 files, 557.5KB)
  Poured from bottle using the formulae.brew.sh API on 2023-12-07 at 10:54:28
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/e/expat.rb
License: MIT
==> Options
--HEAD
	Install HEAD version
==> Caveats
expat is keg-only, which means it was not symlinked into /usr/local,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

If you need to have expat first in your PATH, run:
  echo 'export PATH="/usr/local/opt/expat/bin:$PATH"' >> /Users/turany/.bash_profile

For compilers to find expat you may need to set:
  export LDFLAGS="-L/usr/local/opt/expat/lib"
  export CPPFLAGS="-I/usr/local/opt/expat/include"

For pkg-config to find expat you may need to set:
  export PKG_CONFIG_PATH="/usr/local/opt/expat/lib/pkgconfig"
==> Analytics
install: 21,444 (30 days), 60,563 (90 days), 185,804 (365 days)
install-on-request: 274 (30 days), 841 (90 days), 3,264 (365 days)
build-error: 1 (30 days)
`

fuhlig1: If I remember correctly Geant4 will pick up the expat from the system which doesn't work with Geant4. Unfortunately I can't remember the underlying problem.

dennisklein: Ok, the `expat` from brew is keg-only (not seen by CMake), however, there were changes in FindEXPAT in CMake 3.27+ that broke the detection of expat in Geant4 11.1: https://bugzilla-geant4.kek.jp/show_bug.cgi?id=2556. Geant4 11.2 fixes it and since we use it now, the workaorund is no longer needed.

dennisklein:
Now FairSoft builds again on the macos CI, but there are build and configure errors for FairRoot. Will have a look today.

cmake/legacy.cmake
fuhlig1: Thanks. I think that was the problem. I will try when the RC branch is ready for test.

dennisklein:
But also we do not have xcode 15 yet in the macos CI.

cmake/legacy.cmake
dennisklein: @fuhlig1 Try building the dev branch now. There may potentially be some problems to be solved on FairRoot side however.

dennisklein: I am also in the process of updating the FairSoft (and FairRoot later, maybe only in January) CI images. If problems arise there with new compilers etc, more PRs will follow.

PR #516, opened 2023-12-22
[legacy] ci: Test C++20

dennisklein:
Runs the Fedora 39 test build with `-std=c++20`.


ChristianTackeGSI:
As you change pythia6 here…

can we please open some issue on FairRoot and/or FairSoft about our future intentions regarding Pythia6? So that we can track it?

Just should be referenced here. No need to reference in the commits.

dennisklein:
There is a dedicated pr for pythia6, see #515. I just based this branch on top of it so the legacy CI path is triggered.

dennisklein:
> can we please open some issue on FairRoot and/or FairSoft about our future intentions regarding Pythia6? So that we can track it?

Go ahead and create an issue. I would say this is more of a FairRoot topic.

dennisklein:
Obsoleted by #523

PR #522, opened 2024-01-19
[legacy] Add new package dependencies where missing

fuhlig1:
Add libzst, liblzma and corresponing developer packages in all setup scripts where they were not yet used. 
These two libs are needed to enable boost streaming with compression.

### DIFF
--- a/legacy/setup-archlinux.sh
+++ b/legacy/setup-archlinux.sh
@@ -10,5 +10,5 @@ pacman -S --noconfirm \
     lsb-release lz4 m4 make mesa ncurses net-tools openblas openssl patch \
     pkg-config procps python python-numpy python-pygments python-pyyaml \
     readline rsync sed sqlite subversion tar tbb unzip util-linux-libs wget \
-    which xerces-c xz yaml-cpp
+    which xerces-c xz yaml-cpp zstd
 pacman -Scc --noconfirm
--- a/legacy/setup-debian.sh
+++ b/legacy/setup-debian.sh
@@ -9,9 +9,9 @@ apt-get -y install autoconf automake binutils \
 	hostname libbz2-dev libcurl4-openssl-dev libgsl-dev libicu-dev \
 	libfftw3-dev \
 	libgl1-mesa-dev libglu1-mesa-dev libgrpc++-dev \
-	libncurses-dev libreadline-dev libsqlite3-dev libssl-dev libtool \
+	liblzma-dev libncurses-dev libreadline-dev libsqlite3-dev libssl-dev libtool \
 	libx11-dev libxerces-c-dev libxext-dev libxft-dev \
-	libxml2-dev libxmu-dev libxpm-dev libyaml-cpp-dev lsb-release make patch \
+	libxml2-dev libxmu-dev libxpm-dev libyaml-cpp-dev libzstd-dev lsb-release make patch \
 	python3-dev protobuf-compiler-grpc rsync sed subversion tar unzip wget xutils-dev xz-utils
 apt-get -y -t buster-backports install cmake || \
 	apt-get -y install cmake
--- a/legacy/setup-fedora.sh
+++ b/legacy/setup-fedora.sh
@@ -9,7 +9,7 @@ dnf -y install \
     grpc-plugins gsl-devel gzip help2man hostname hwloc-devel libX11-devel libXau-devel \
     libXdmcp-devel libXext-devel libXfont-devel libXft-devel libXmu-devel libXpm-devel \
     libXrender-devel libbsd-devel libicu-devel libjpeg-turbo-devel libtiff-devel libtool \
-    libunistring-devel libuuid-devel libxml2-devel lz4-devel m4 make make mesa-libGL-devel \
+    libunistring-devel libuuid-devel libxml2-devel libzstd-devel lz4-devel m4 make make mesa-libGL-devel \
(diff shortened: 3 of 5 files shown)


dennisklein (approved):
Open todo: rebuild the CI images

rbx (approved):

fuhlig1:
On macosx the added libraries weren't found. They are installed using brew into non system paths. These are not found by boost automatically such that we need to give boost some hints where to find them. This is done with the second commit.
Please have a look if the code is acceptable.

dennisklein:
> On macosx the added libraries weren't found. They are installed using brew into non system paths. These are not found by boost automatically such that we need to give boost some hints where to find them. This is done with the second commit. Please have a look if the code is acceptable.

Hm, that is odd. Looking at brew formulas they are *not keg-only*, meaning they should be installed into a system prefix.

Can you check with @rbx if this is also needed on our CI macs? Sounds a bit like maybe an artifact from brew changing the system prefix over time or something (/opt vs /usr, wasn't there soemthing?)?

rbx:
All three CI macs install these under `brew --prefix`, but that is not always the same:

MacOS 14.2 (m1): `/opt/homebrew`
MacOS 14.2 (x86): `/usr/local`
MacOS 13.6 (x86): `/usr/local`

Did we ever change these manually?

dennisklein:
Looks like default homebrew prefix on intel macs is `/usr/local` which I would expect is searched by default. On Apple silicon homebrew default prefix is `/opt/homebrew` - here I expect that the homebrew installation procedure has do register this prefix as a system prefix somehow. Can we check how this is done and whether this config is present on the machines where you need pass the installation paths explicitely?

fuhlig1:
That is exactly the point. On Intel boost finds the two libraries from /usr/local. I overlooked the point since I developed only on macosx. I will check if /opt/homebrew can easily be added as system path, at least for the fairsoft build on macosx on Arm.

rbx:
> That is exactly the point. On Intel boost finds the two libraries from /usr/local. I overlooked the point since I developed only on macosx. I will check if /opt/homebrew can easily be added as system path, at least for the fairsoft build on macosx on Arm.

For fairsoft, or maybe just for boost alone

fuhlig1:
Only for boost alone but I couldn't find yet anything how to add this to the boost build system globally, that's why I added the needed information for the two libraries explicitly.

rbx:
maybe with this:
https://www.bfgroup.xyz/b2/manual/release/index.html#bbv2.builtin.features.library-path

dennisklein:
https://www.bfgroup.xyz/b2/manual/release/index.html#bbv2.tasks.projectsearch.path

fuhlig1:
Okay. I think I found the solution. What I did explicitly for the two packages obviously works globally for boost. I will update the PR soon.

dennisklein:
I still wonder how do the other packages in FairSoft find stuff in `/opt/homebrew`? (Do they?) Do we understand this now?

dennisklein:
> I still wonder how do the other packages in FairSoft find stuff in `/opt/homebrew`? (Do they?) Do we understand this now?

I think CMake infers some search paths from binaries it finds in `$PATH`?!

fuhlig1:
Update. I can set the paths when executing b2 such that boost find both packages. Unfortunately this then breaks the build. Some ipp file within boost isn't any longer found. So adding the paths globally doesn't work. 
If someone finds a better solution please let me know, otherwise I would propose the current solution which we somehow also use to pass the correct python version to boost.

FairSoftConfig.cmake
```diff
@@ -79,3 +79,79 @@ if(APPLE)
...
+    message(WARNING "LZMA installation not found. Boost iostream will be build without lzma compression")
+    set(LZMA_EXTERNAL "" CACHE STRING "LZMA exetrnal usage for Boost" FORCE)
+    set(LZMA_VERSION "" CACHE STRING "LZMA version info for Boost" FORCE)
+    set(LZMA_STRING "" CACHE STRING "LZMA install info for Boost" FORCE)
+  endif()
+else()
+  set(LZMA_EXTERNAL "" CACHE STRING "LZMA exetrnal usage for Boost" FORCE)
+  set(LZMA_VERSION "" CACHE STRING "LZMA version info for Boost" FORCE)
+  set(LZMA_STRING "" CACHE STRING "LZMA install info for Boost" FORCE)
+endif()
```
dennisklein, line 121: It is not guaranteed that the user will use `-C FairSoftConfig.cmake`.

The dependency on `brew` should stay in this file to detect the lzma/zstd prefix and set the standard CMake prefix hints (`LIBLZMA_ROOT` and `LIBZSTD_ROOT`). The rest of the logic (version extraction) should go to `cmake/legacy.cmake` (conditionally on whether the prefix was passed in).

The msg output is not really correct. The script checks, if there is a brew installation of lzma and if not, it prints "Boost iostream will be build without lzma compression". How do you know this at this point? There can be other lzma installations present in the system?

fuhlig1: @dennisklein,

I put everything to cmake/legacy.cmake, please check the new MR #528. For LZMA there is a standard find module. For ZSTD I add one. CMake will try to find both packages independent of the system but will add the necessary information to the boost build system in case of macosx. I only tested the logic so far on my Apple silicon but will test it now also for my Intel based Mac and Debian 10.

dennisklein:
obsolete

PR #523, opened 2024-01-19
Bump DDS to 3.8

rbx:
I also asked @AnarManafov to look at the applied patches, to see if some of them can be removed with this release.

But let's see what CI says for this.


AnarManafov:
I checked legacy patches, which were used for 3.6. 
The ones for the code, we can remove now. Regarding the other two, for cmake configuration. I am not sure about them. I would exclude them for now and see how it builds the matrix.

rbx:
On Ubuntu 20.04 no recent enough protobuf is available:
```
Could NOT find Protobuf: Found unsuitable version "3.6.1", but required is at least "3.15"
```
On Rockylinux-8 no Protobuf is found at all, which is just not installed in that image.

MacOS compiles fine but FairRoot test build fails as we've seen before (not DDS related).

rbx:
@fuhlig1 Do we care about the Ubuntu 20.04 failure? We can either:
- build our own protobuf there.
- not build DDS for this OS.
- not build DDS.
- remove 20.04.

fuhlig1:
> @fuhlig1 Do we care about the Ubuntu 20.04 failure? We can either:
> 
> * build our own protobuf there.
> * not build DDS for this OS.
> * not build DDS.
> * remove 20.04.

This is something I can't answer easily but my proposal would be to remove Ubuntu 20.04 from the build matrix. With Ubuntu 22.04 there is another LTS version and the next will LTS version will appear soon. I am also not aware of any experiment using Ubuntu 20.04. If I am wrong with the last statement please correct me.

cmake/legacy.cmake
```diff
@@ -161,16 +161,12 @@ if(ICU_ROOT)
   set(dds_icu_hint "-DDDS_LD_LIBRARY_PATH=${ICU_ROOT}/lib")
 endif()
 list(APPEND packages dds)
-set(dds_version "3.6")
+set(dds_version "3.8")
 ExternalProject_Add(dds
   GIT_REPOSITORY https://github.com/FairRootGroup/DDS GIT_TAG ${dds_version}
   ${CMAKE_DEFAULT_ARGS} CMAKE_ARGS
     "-DBoost_NO_BOOST_CMAKE=ON"
```
dennisklein, line 168:
```suggestion
```

Can we try without this line, it should not be needed any more.

rbx: commited.

dennisklein (changes requested):
* Then let's remove Ubuntu 20.04 (Jenkinsfile and Docs)
* Please cherry-pick [`ffb98c5` (#516)](https://github.com/FairRootGroup/FairSoft/pull/516/commits/ffb98c598b17381d0a59735e749394080942bb4d) and [`79060ff` (#516)](https://github.com/FairRootGroup/FairSoft/pull/516/commits/79060ff64fb9fa9acb42e31d56a105fc39c3d3da)
* Add the new protobuf dependency of DDS to the `legacy/setup*.sh` scripts

rbx:
done

rbx:
All builds succeed now. MacOS stuck on the same FairRoot test.

dennisklein (approved):

rbx:
Fedora 39 also runs, which is set to C++20 now.

PR #524, opened 2024-01-29
[legacy] vc: Bump 1.4.4

dennisklein:


rbx (approved):

rbx:
actually, vc download failed. hiccup or something wrong? I'll rerun.

dennisklein:
Hm, maybe url is wrong.

cmake/legacy.cmake
```diff
@@ -242,10 +242,10 @@ ExternalProject_Add(hepmc
 )
 
 list(APPEND packages vc)
-set(vc_version "1.4.3")
+set(vc_version "1.4.4")
 ExternalProject_Add(vc
   URL https://github.com/VcDevel/Vc/releases/download/${vc_version}/Vc-${vc_version}.tar.gz
```
rbx, line 247:
```suggestion
  URL https://github.com/VcDevel/Vc/archive/refs/tags/${vc_version}.zip
```
🤷‍♂️

rbx (changes requested):

cmake/legacy.cmake
rbx: There's only the source code link for this version: https://github.com/VcDevel/Vc/releases/tag/1.4.4
Can we build from that?

rbx: Seems to be the same thing though.

rbx: I've tested this locally and it builds. I'll commit to see how it does on the CI. Can drop if anything is wrong.

rbx (approved):

PR #528, opened 2024-02-12
[legacy] boost: find needed libs on macosx

fuhlig1:
When brew installs software to a directory other than /usr/local boost can't find the packages since they are not in the default search paths. This is the case on Apple silicon. Let CMake find the packages and pass the proper information to the boost build process.

### DIFF
--- a/cmake/FindZSTD.cmake
+++ b/cmake/FindZSTD.cmake
@@ -1,0 +1,60 @@
+################################################################################
+# Copyright (C) 2024 GSI Helmholtzzentrum fuer Schwerionenforschung GmbH       #
+#                                                                              #
+#              This software is distributed under the terms of the             #
+#              GNU Lesser General Public Licence (LGPL) version 3,             #
+#                  copied verbatim in the file "LICENSE"                       #
+################################################################################
+#[=======================================================================[
+FindZSTD
+-----------
+
+Find ZSTD compression algorithm headers and library.
+
+Result variables
+^^^^^^^^^^^^^^^^
+
+This module will set the following variables in your project:
+
+``ZSTD_FOUND``
+  True if libzstd headers and library were found.
+``ZSTD_INCLUDE_DIR``
+  Directory where libzstd headers are located.
+``ZSTD_LIBRARY``
+  Zstd library to link against.
+``ZSTD_BINARY``
+  Zstd binary
+``ZSTD_VERSION``
+    the version of ZSTD found.
+#]=======================================================================]
+
+find_path(ZSTD_INCLUDE_DIR zstd.h)
+
+if(NOT ZSTD_LIBRARY)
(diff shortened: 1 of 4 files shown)


cmake/legacy.cmake
```diff
@@ -6,6 +6,50 @@
...
 ################################################################################
 cmake_minimum_required(VERSION 3.19...3.28 FATAL_ERROR)
+cmake_policy(VERSION 3.19...3.28)
+
+
+if(APPLE)
+  execute_process(COMMAND brew --prefix OUTPUT_VARIABLE brew_prefix)
+  set(LibLZMA_ROOT ${brew_prefix})
+  set(ZSTD_ROOT ${brew_prefix})
+endif()
```
dennisklein, line 16: Move this please to the `FairSoftConfig.cmake` file (which is mostly needed on macs right now). The idea is that the only this file adds a dependency to `brew` and not the general `cmake/legacy.cmake` file.

dennisklein (changes requested):
Looks very nice, one small change, see my inline comment.

cmake/legacy.cmake
fuhlig1: Fixed with the last commit.

dennisklein (approved):

dennisklein:
let's wait for the CI and then we can merge from my point of view

cmake/FindZSTD.cmake
```diff
@@ -0,0 +1,48 @@
+#[=======================================================================[.rst:
+FindZSTD
+-----------
```
ChristianTackeGSI, line 3: Please excuse me for being nit picky…

Where is this file coming from? What's the copyright?

Or do I miss something?

cmake/legacy.cmake
```diff
@@ -6,6 +6,43 @@
 #                  copied verbatim in the file "LICENSE"                       #
 ################################################################################
```
ChristianTackeGSI, line 7: Please update the copyright. I think, we already have 2024.

cmake/FindZSTD.cmake
fuhlig1: @ChristianTackeGSI,

the file does not come from somewhere but was inspired by FindLibLZMA.cmake, so I took the same documentation block used by official  modules. 
You are right, I should add our copyright.

cmake/legacy.cmake
fuhlig1: Will be updated.

cmake/FindZSTD.cmake
fuhlig1: Fixed with last commit. 
Also add extracting version info in case zstd binary is not available.

ChristianTackeGSI (approved):
Please wait for the CI (it it's not ready anyway). :-)

PR #531, opened 2024-02-29
[legacy] Address latest issues reported in #529

### LINKED ISSUE #529: FairSoft jan24 on Debian10
The required cmake version for jan24 is at least 3.19 but with buster-backports on Debian 10 you only get 3.18. Thus the cmake stage of jan24 stops with:
```
CMake Error at cmake/legacy.cmake:8 (cmake_minimum_required):
  CMake 3.19 or higher is required.  You are running version 3.18.4
Call Stack (most recent call first):
  CMakeLists.txt:32 (include)
  ```
It would be great if a fix for this problem could be provided.

### PR DESCRIPTION
**Pull request by dennisklein**
fix #529

### DIFF
--- a/legacy/setup-debian.sh
+++ b/legacy/setup-debian.sh
@@ -3,16 +3,14 @@
 export DEBIAN_FRONTEND=noninteractive
 apt-get update
 apt-get -y upgrade
-apt-get -y install autoconf automake binutils \
-	bison build-essential bzip2 ca-certificates coreutils \
-	curl debianutils file findutils flex g++ gcc gfortran git gzip \
-	hostname libbz2-dev libcurl4-openssl-dev libgsl-dev libicu-dev \
-	libfftw3-dev \
-	libgl1-mesa-dev libglu1-mesa-dev libgrpc++-dev \
-	liblzma-dev libncurses-dev libreadline-dev libsqlite3-dev libssl-dev libtool \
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: legacy/setup-debian.sh [hunk 1]
```diff
@@ -3,16 +3,14 @@
...
-	hostname libbz2-dev libcurl4-openssl-dev libgsl-dev libicu-dev \
-	libfftw3-dev \
-	libgl1-mesa-dev libglu1-mesa-dev libgrpc++-dev \
-	liblzma-dev libncurses-dev libreadline-dev libsqlite3-dev libssl-dev libtool \
-	libx11-dev libxerces-c-dev libxext-dev libxft-dev \
-	libxml2-dev libxmu-dev libxpm-dev libyaml-cpp-dev libzstd-dev lsb-release make patch \
-	python3-dev protobuf-compiler-grpc rsync sed subversion tar unzip wget xutils-dev xz-utils
-apt-get -y -t buster-backports install cmake || \
-	apt-get -y install cmake
+apt-get -y install autoconf automake binutils bison build-essential bzip2 \
```
fuhlig1, line 6: I would propose to put only one package per line. In the diff one cannot find the one added package. With the proposed change this is true for this commit but will not be any problem in later changes.

dennisklein: That's just a limitation of github :P 

Check git's `--word-diff` option:

image

dennisklein:
obsoleted by #540

PR #534, opened 2024-04-02
[legacy] root: Bump v6.30.04

dennisklein:
Get in latest ROOT fixes and disable `runtime_cxxmodules` as those are breaking frequently on many platforms.


fuhlig1 (approved):

ChristianTackeGSI:
Note that https://github.com/root-project/root/issues/9815 is still not resolved. So people will not get deprecation warnings in the cling interpreter.

(Not saying we should enable `runtime_cxxmodules`. Just a note, so that we know about this pending thing.)

PR #543, opened 2024-06-28
fix URLs for HepMC tarball

jezwilkinson:
following a server migration on hepmc.web.cern.ch, the URI no longer has the "/hepmc/" subdirectory (both for the homepage and the releases/ folder)


dennisklein (approved):
I will have a look at the CI failure, it seems to be an effect of the recent upgrade from Singularity 2.x to Apptainer 1.3 on the CI boxes.

PR #555, opened 2024-10-22
Check for system installation of lohmann-json

fuhlig1:
Switch off the root builtin version if a system installation exist. On macosx there is a compilation error otherwise.

### DIFF
--- a/cmake/legacy.cmake
+++ b/cmake/legacy.cmake
@@ -374,13 +374,19 @@ else()
   unset(root_cocoa)
   set(root_x11 ON)
 endif()
+find_package(nlohmann_json 3.9)
+if(nlohmann_json_FOUND)
+  set(root_builtin_nlohmannjson "-Dbuiltin_nlohmannjson=OFF")
+else()
+  set(root_builtin_nlohmannjson "-Dbuiltin_nlohmannjson=ON")
+endif()
+
 ExternalProject_Add(root
   GIT_REPOSITORY https://github.com/root-project/root/ GIT_TAG v${root_version_gittag}
   GIT_SHALLOW 1
   ${CMAKE_DEFAULT_ARGS} CMAKE_ARGS
     "-Daqua=ON"
     "-Dasimage=ON"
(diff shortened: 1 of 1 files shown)


cmake/legacy.cmake
```diff
@@ -370,17 +370,24 @@ endif()
...
+  set(lohmann_json "-Dbuiltin_nlohmannjson=ON")
 endif()
 ExternalProject_Add(root
   GIT_REPOSITORY https://github.com/root-project/root/ GIT_TAG v${root_version_gittag}
   GIT_SHALLOW 1
   ${CMAKE_DEFAULT_ARGS} CMAKE_ARGS
     "-Daqua=ON"
     "-Dasimage=ON"
-    "-Dbuiltin_nlohmannjson=ON"
+    ${lohmann_json}
```
dennisklein, line 390:
```suggestion
    ${root_builtin_nlohmannjson}
```

and sort it together with the existing `root_builtin_glew` variable towards the end of the option list for consistent variable naming, thx!

cmake/legacy.cmake
```diff
@@ -370,17 +370,23 @@ endif()
 if(APPLE)
   set(root_cocoa "-Dcocoa=ON")
   set(root_x11 OFF)
+  find_package(nlohmann_json 3.9)
```
ChristianTackeGSI, line 373: Why only on APPLE?
Yes, you want to fix the problem on that arch.

But if we have a good enough system package on Linux, why not use it?

fuhlig1: Because it is/was working on the other systems without problems. ;-}}
I did not want to introduce new problems on the other supported systems, but I am also fine to add the same logic for other systems.

ChristianTackeGSI: Just move the logic out of the `if(APPLE)`?

… and let the CI do its job. ;-)

Let's keep our code simple, if possible?

ChristianTackeGSI (approved):
LGTM

fuhlig1:

PR #556, opened 2024-10-22
Fix macosx opengl

fuhlig1:
With new versions of the OS and/or the command line tools a OpenGL version
installed with brew was used which somehow doesn't work. With the change the
system version from the OpenGL framework is used.


dennisklein (approved):

fuhlig1:

PR #557, opened 2024-10-23
Build ROOTs Pythia8 interface

fuhlig1:
The default setting for the pythia8 interface was switched from ON to OFF
such that we have to switch it on explicitly.

### DIFF
--- a/cmake/legacy.cmake
+++ b/cmake/legacy.cmake
@@ -400,6 +400,7 @@ ExternalProject_Add(root
     "-Dmlp=ON"
     "-Dpyroot=ON"
     "-Dpythia6=ON"
+    "-Dpythia8=ON"
     "-Dreflex=OFF"
     "-Droofit=ON"
     "-Druntime_cxxmodules=OFF"


dennisklein:
* Can you please add the link to the commit or PR in the root github repo, that switched the default to the commit msg body.
* Also, please prefix the first line in the commit msg with `[legacy] `, thx!

fuhlig1:
@dennisklein,

is there a way to block starting the CI since the last run was successful and the changes are only in the commit message?

fuhlig1:
Ok, then I will stop the run on Jenkins.

dennisklein:
Ah, I remember, @ChristianTackeGSI added a short delay of a minute or two. So, if you are fast enough, the ci jobs should not even start, but only show up in the jenkins job queue in the left sidebar.

fuhlig1:
Unfortunately I didn't find the job before it starts. Anyway I stopped it now from Jenkins.

dennisklein:
Please add the full github commit url to the commit msg. It will then render as a clickable link, see for example https://github.com/FairRootGroup/FairSoft/commit/5e5d0c41bab0b2dfd153898046e2bff302b3632b

dennisklein:
Also, github is then able to render a link in the activity stream of the root issue/PR, if there is a fitting one. For a commit link, it will only work in one direction, which is nice already, of course.

fuhlig1:
Done. I also found the Jenkins build queue to remove the build before it was started.

dennisklein (approved):

fuhlig1:

PR #559, opened 2024-11-15
Update nov22 patches branch

fuhlig1:
Allow compilation with Ubuntu 24.04 and Fedora 41. Also test OpenSuse 15.5 as new system.
Confirm that the changes are also working with already tested systems Debian 10 and Debian 12.

Did not change any package versions, only add fixes as patches. This version should be then tagged as nov22p3.


cmake/legacy.cmake
```diff
@@ -130,11 +130,17 @@ set(boost_features
   "visibility=hidden"
   "pch=off"
 )
+
+if(${Python_VERSION} VERSION_GREATER_EQUAL 3.13.0)
+  set(python_3_13_fix COMMAND ${patch} -p0 -i "${CMAKE_SOURCE_DIR}/legacy/boost/PyEval_CallMethod_fix_python_3-13.patch")
+endif()
+
```
ChristianTackeGSI, line 137: Is this patch really harmful on earlier Python releases?

I doubt that, as [`PyObject_CallMethod` exists for a long time now and was last changed in Python 3.4](https://docs.python.org/3/c-api/call.html#c.PyObject_CallMethod) (Debian-10 has 3.7).

If boost uses this patch conditionally, then I am fine.

Otherwise please patch unconditionally.

legacy/README.md
```diff
@@ -94,11 +94,14 @@ please contact us.
...
 | Fedora     | x86_64 | 36    | GCC 12.2.1                 | 3.24.2 |
 | Fedora     | x86_64 | 37    | GCC 12.2.1                 | 3.24.2 |
+| Fedora     | x86_64 | 41    | GCC 14.2.1                 | 3.30.5 |
 | macOS      | x86_64 | 12    | AppleClang 13, gfortran 12 | 3.24.1 (brew) |
 | macOS      | x86_64 | 11    | AppleClang 13, gfortran 12 | 3.24.1 (brew) |
 | macOS      | arm64  | 12    | AppleClang 13, gfortran 12 | 3.24.1 (brew) |
+| OpenSuSe   | x86_64 | 15.5  | GCC 7.5.0                  | 3.20.4 |
 | Ubuntu     | x86_64 | 20.04 | GCC 9.4.0                  | 3.24.3 (`bootstrap-cmake.sh`) |
 | Ubuntu     | x86_64 | 22.04 | GCC 11.3.0                 | 3.22.1 |
+| Ubuntu     | x86_64 | 24.04 | GCC 13.2.0                 | 3.28.3 |
```
ChristianTackeGSI, line 104: If you claim, that those are tested, then please
1. Add them to the container build rules: https://github.com/FairRootGroup/FairSoft/tree/master/test/container/legacy
2. Ping @dennisklein to rebuild the CI containers
3. Add them to Jenkinsfile for actual testing
4. Then add to README.

I'd suggest to perform all that in another PR round.

cmake/legacy.cmake
```diff
@@ -265,13 +273,21 @@ set(clhep_version "2.4.5.1")
...
+if(${CMAKE_VERSION} VERSION_LESS 3.27.0)
+  set(patch_step_name patch)
+else()
+  set(patch_step_name patch_disconnected)
+endif()
+
 set(clhep_source ${CMAKE_BINARY_DIR}/Source/clhep)
-ExternalProject_Add_Step(clhep move_dir DEPENDEES download DEPENDERS patch
+ExternalProject_Add_Step(clhep move_dir DEPENDEES download DEPENDERS ${patch_step_name}
```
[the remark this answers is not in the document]
fuhlig1: Done.

cmake/legacy.cmake
```diff
@@ -351,6 +368,8 @@ ExternalProject_Add(root
   COMMAND ${patch} -p1 -i "${CMAKE_SOURCE_DIR}/legacy/root/fix_tpython_python_3_11_deprecation.patch"
   COMMAND ${patch} -p1 -i "${CMAKE_SOURCE_DIR}/legacy/root/fix_pyroot_cast_error_python_3_11.patch"
   COMMAND ${patch} -p1 -i "${CMAKE_SOURCE_DIR}/legacy/root/fix_root_install_external_tars_with_cmake_3_24.patch"
+  COMMAND ${patch} -p1 -i "${CMAKE_SOURCE_DIR}/legacy/root/fix_compilation_with_gcc13.patch"
+  COMMAND ${patch} -p1 -i "${CMAKE_SOURCE_DIR}/legacy/root/fix_strlcpy_if_glibc_gt_2.38.patch"
```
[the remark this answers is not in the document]
fuhlig1: Yes I did consider this but finally decided against this solution. The simply reason for the decision was that CBM did a large check based on this version. Bumping ROOT may, as you already mentioned, introduce new problems. Bumping would also mean for CBM to recheck everything for the release which I wanted to avoid.

cmake/legacy.cmake
fuhlig1: Probably the patch isn't harmful for older versions of Python releases, but I didn't want to introduce new problems when trying to fix new ones such that I decide to do the patch conditionally.
I will check what is done by the boost project themself.

fuhlig1: Since boost itself uses the patch unconditionally I changed it in the same way.

legacy/README.md
fuhlig1: I overlooked the the line claiming that the  systems are tested regularly. This is for sure not the case. They were only tested once. I agree to remove them from this list, but there is still the question if we want to add information about systems which were only tested once.

fuhlig1: Now I wonder. Looking at the build and test matrix for this PR I see that only 4 systems are tested. How do we test the rest of the list regularly?

ChristianTackeGSI: As already explained privately, we used to have different test sets for Pull Requests and for "branch builds". So when master or any branch gets an update, a bigger test suite is running.

The idea is to have "faster responses" on PRs and the full coverage on the main braanches.

Github lately introduced a "patch queue", which we could use for a middle thing: Let the initial testing for a PR be quick. When we consider the thing to be "good enough", we can press the "add to merge queue" and then a full test round is running. If it is successful, it's merged and done. Otherwise it's back to square one.

Just ponedering. As also already said privately–I am already quite far away from all of this.

cmake/legacy.cmake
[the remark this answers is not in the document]
ChristianTackeGSI: Consider documenting this in the git commit body.

legacy/README.md
fuhlig1: I have added a new section about systems not yet in the test matrix but tested manually.

PR #568, opened 2025-02-19
RC mar25

fuhlig1:
Bump versions of many packages.
Remove pythia6.
Add VecGeom as new package.


dennisklein:
Went through all commits and added updates to the README table as well as removed the actual patch files in cases where the patches were no longer applied.

Also bumped ROOT to 6.24.08 for latest mac support.

fuhlig1:
Outdated. Close the issue since the packages were updated to newer versions already in the dev branch.

PR #587, opened 2026-07-07
Fix build on macosx26

fuhlig1:
The PR fixes the following two problems:
- Use correct new download URL for Pythia8 which is needed for all supported platforms
- Backport fixes for ROOT such that Root v6.36.08 can be compiled with macos SDK 26.5


fuhlig1:
The four commits need to be ported to the feb26_patches branch.
The commit fixing the pythia8 download URL need to be ported to the nov24_patches branch.

karabowi (approved):

